PR #4, opened 2019-07-24
Feature/navigation item

jvhoven:


packages/core/FlashMessage/src/FlashMessage.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * FlashMessage is a graphical implementation for @see {@link https://github.com/42BV/react-flash-messages}.
+ *
+ * Use it when you want to globally show a notification / message.
+ *
+ * @exports FlashMessage
+ * @returns {JSX.Element}
+ */
+export default function FlashMessage({ className }: Props): JSX.Element {
```
MrHus: Is `React.ReactNode` not a better return type?

PR #194, opened 2019-11-07
feat: Added SearchInput component.

### LINKED ISSUE #191: SearchInput component
Introduce `SearchInput` where `onChange` will be debounced until a user is done typing.

- [ ] Add `debounce` property on `Input`.
- [ ] Create `SearchInput`.

MrHus:
The `SearchInput` component is a small wrapper around the `reactstrap`
`Input` component which debounces the `onChange` by `500` milliseconds
by default. It uses `lodash`'s `debounce` under the hood.

Features:

  1. When the user presses enter the `SearchInput` will skip the debounce
     and set the value straight away.

  2. The icon which is a magnifying class can be toggled on or of.

  3. The debounce timeout can also be set to a custom value.

  4. The debounce settings which lodash accepts can also be set for
     maximum configurability.

The modal pickers now use the `SearchInput` instead of debouncing the
inputs themselves.

Also had to set the border color in `.form-control:focus` to a gray
to prevent it from becoming blue for the `SearchInput`.

Closes #191

### DIFF
--- a/src/core/SearchInput/SearchInput.test.tsx
+++ b/src/core/SearchInput/SearchInput.test.tsx
@@ -1,0 +1,150 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
+
+import lodash from 'lodash';
+
+import SearchInput from './SearchInput';
+import { Input } from 'reactstrap';
+
+describe('Component: SearchInput', () => {
+  function setup({
+    debounce,
+    showIcon,
+    debounceSettings
+  }: {
+    debounce?: number;
+    showIcon: boolean;
+    debounceSettings?: lodash.DebounceSettings;
+  }) {
+    // @ts-ignore
+    jest.spyOn(lodash, 'debounce').mockImplementation(fn => {
+      return fn;
+    });
+
+    const onChangeSpy = jest.fn();
+
+    const searchInput = shallow(
+      <SearchInput
+        value=""
+        debounce={debounce}
+        debounceSettings={debounceSettings}
+        onChange={onChangeSpy}
+        showIcon={showIcon}
+      />
+    );
+
+    return { searchInput, onChangeSpy };
+  }
+
+  describe('ui', () => {
+    test('with icon', () => {
+      const { searchInput } = setup({ showIcon: true });
+
+      expect(toJson(searchInput)).toMatchSnapshot();
+    });
+
+    test('with icon when undefined', () => {
+      const { searchInput } = setup({ showIcon: undefined });
+
+      expect(toJson(searchInput)).toMatchSnapshot();
+    });
+
+    test('without icon', () => {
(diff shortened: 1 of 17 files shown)


src/form/ModalPicker/multiple/ModalPickerMultiple.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { FormGroup, Label, Input } from 'reactstrap';
-import { debounce, isArray } from 'lodash';
+import { isArray } from 'lodash';
```
jvhoven: `Array.isArray` could be an alternative which is, in my opinion, preferred.

src/core/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+    onChange,
+    showIcon = true,
+    ...rest
+  } = props;
+
+  const handleChange = useRef(
+    lodashDebounce(
+      (value: string) => {
+        onChange(value);
+      },
```
jvhoven:
```
lodashDebounce(onChange, debounce, debounceSettings)
```

src/core/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+      debounce,
+      debounceSettings
+    )
+  );
+
+  function handleKeyUp(event: KeyboardEvent<HTMLInputElement>) {
+    if (event.key === 'Enter') {
+      // @ts-ignore
+      onChange(event.target.value);
```
jvhoven:
```suggestion
      onChange(event.currentTarget.value);
```

To eliminate the need for the `@ts-ignore`.

src/core/SearchInput/SearchInput.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    describe('onKeyUp behavior', () => {
+      it('should on "ENTER" press set the value immediately', () => {
+        const { searchInput, onChangeSpy } = setup({ showIcon: true });
+
+        searchInput
+          .find(Input)
+          .props()
+          // @ts-ignore
+          .onKeyUp({ key: 'Enter', target: { value: 'Maarten' } });
```
jvhoven:
```suggestion
          .onKeyUp({ key: 'Enter', currentTarget: { value: 'Maarten' } });
```

To eliminate the need for a `@ts-ignore` within `SearchInput.tsx`

src/core/SearchInput/SearchInput.tsx
MrHus: Good catch, leftover from when I did more inside the lambda.

src/core/SearchInput/SearchInput.tsx
MrHus: Did not know this, good to know 👍

PR #205, opened 2019-11-12
fix: select placeholder is now a selectable option

### LINKED ISSUE #137: `Select`: placeholder should be a selectable option
Currently its a non-selectable option which in turn shows the first item in the options as selected even though its value is not selected. This is incorrect behaviour. Making the placeholder a selectable option would resolve this.

jvhoven:
In the previous version of `Select` the placeholder was a non-selectable
option which in turn shows the first item in the options as selected,
even though it is not selected. This is resolved by making the
placeholder a selectable option.

~Additionally this change contains a fix for object equality when passing
a value to `Select`. This is done by using `Lodash.isEqual` instead of
trying to do equal comparison.~

Fixes #137.

### DIFF
--- a/src/form/Select/Select.tsx
+++ b/src/form/Select/Select.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { FormGroup, Label, Input as RSInput } from 'reactstrap';
-import { get, constant, isArray } from 'lodash';
+import { get, constant } from 'lodash';
 import { InputType } from 'reactstrap/lib/Input';
 import { Page } from '@42.nl/spring-connect';
 
@@ -114,7 +114,7 @@ export default class Select<T> extends Component<Props<T>, State<T>> {
 
     const { options } = props;
 
-    if (isArray(options)) {
+    if (Array.isArray(options)) {
       this.state = {
         options,
         loading: false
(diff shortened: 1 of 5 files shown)


src/form/Select/Select.tsx
```diff
@@ -221,12 +221,11 @@ export default class Select<T> extends Component<Props<T>, State<T>> {
       className: value === undefined ? 'showing-placeholder' : ''
     };
 
-    // @ts-ignore
-    const indexOfValue = options.findIndex(option => option === value);
+    const indexOfValue = options.findIndex(option => isEqual(option, value));
```
MrHus: I think `useEqual` is a bit of a nuclear weapon pointed at this problem. In all other components we simply check if the `optionForValue` returns the same label for that option. In other words we compare the labels and if they are equal it is the same option.

Created an issue for this: https://github.com/42BV/ui/issues/209
This will not work for every scenario, so I think we should add a `compare` prop to all of our form elements. It should take two T's and return if they are the same.

In most cases this function would be `(a: T, b: T) => a.id === b.id`, then the user can decide they want to go full nuclear with `(a: T, b: T) => isEqual(a, b)`.

The fallback should then be the labels check.

jvhoven: Reverted the check back.

PR #211, opened 2019-11-13
Bugfix/pager can go out of bounds

MrHus:
The bounds of the `Pager` are now constricted by adding a `disabled`
property on the buttons. When `first` is `true` the `previous` button
is disabled. When `last` is true the `next` button is disabled.
    
Since `Button` did not support `disabled` I've added support for it.
The same goes for the `Icon` which now also supports `disabled`.
    
Also fixed a missing unit test for the `Select` component. There was not
test for when the `value` was not longer in the list of `options`.


MrHus:
Hold the presses, new CSS around icons breaks style.

PR #214, opened 2019-11-14
feat: Added IconPicker form component.

### LINKED ISSUE #208: IconPicker component
Would be nice to have a component from which the user can select one of the icons from material design. Useful when applications want the user to be able to select items.

MrHus:
With the `IconPicker` form component the user can select an icon from
the material design icons.

Changed the icons to an array of strings, and create the `IconType` type
from that array of strings. This way we can iterate over the icons and
render them in the IconPicker.

Also removed icons which did not exist in our version of the material icons.
Bumped the material icons to the latest version which is `v49`.

Promoted the `pageOf` test utility to the actual source code. It is used
in the `IconPicker` to create pages from the icons array.

Minor change: fixed jarb comment in `CheckboxMultipleSelect` and formatted
the file.

Closes #208


src/form/IconPicker/IconPicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+        isOpen={isOpen}
+        target="icon-picker-popover"
+        trigger="legacy"
+        toggle={() => setIsOpen(!isOpen)}
+      >
+        <PopoverHeader>
+          <SearchInput value={query} onChange={setQuery} debounce={0} />
+        </PopoverHeader>
+        <PopoverBody>
+          <div className="text-center">
```
MrHus: Width should be consistent when there are few results, so the width should be fixed

src/form/IconPicker/IconPicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+    error,
+    onChange,
+    onBlur
+  } = props;
+
+  const [isOpen, setIsOpen] = useState(false);
+  const [pageNumber, setPageNumber] = useState(1);
+  const [query, setQuery] = useState('');
+
+  const filteredIcons = icons.filter(icon =>
```
MrHus: Add empty state for when the query does not return any icons.

src/form/IconPicker/IconPicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+   * The placeholder of the form element.
+   */
+  placeholder: string;
+}
+
+/**
+ * IconPicker is a form element which allows the user to select one
+ * of the material design icons. It is a popover which shows
+ */
+export default function IconPicker(props: Props) {
```
MrHus: Add ability to clear the icon selection.

src/form/IconPicker/IconPicker.tsx
```diff
@@ -0,0 +1,240 @@
...
+              if (isOpen) {
+                searchInputApi.setValue('');
+              }
+
+              setIsOpen(!isOpen);
+            }}
+          >
+            <PopoverHeader>{searchInput}</PopoverHeader>
+            <PopoverBody>
+              <div style={{ width: 250, height: 250 }}>
```
jvhoven: Bit tricky responsive wise, but I understand if this is not the focus for now.

PR #231, opened 2019-11-22
improvement: EpicTable is now flat when no horizontal scrollbar is shown

### LINKED ISSUE #230: EpicTable improve tables that do not need the horizontal scrollbar
Multiple people are taken aback by the shadows produced on a very small table. This is because they do not see that the center can be scrolled in this case. So when a table is small enough not to need a horizontal scrollbar it should not render the shadows and instead be more flat.

MrHus:
Multiple people are taken aback by the shadows produced on a very small
table. This is because they do not see that the center can be scrolled
in this case. So when a table is small enough not to need a horizontal
scrollbar it should not render the shadows and instead be more flat.

The technical solution is to calculate the total desired width of the
center as given by the user. When the desired width can be met, there is
no need for a horizontal scrollbar. In that case we do not render shadows
around the left and right column, but only around the container.

Also the `GooeyCenter` now renders at 100% width to prevent width
flickering.

Closes #230

### DIFF
--- a/src/table/EpicTable/EpicTable.stories.tsx
+++ b/src/table/EpicTable/EpicTable.stories.tsx
@@ -669,6 +669,69 @@ storiesOf('table|EpicTable', module)
       </Card>
     );
   })
+  .add('small example', () => {
+    return (
+      <Card body>
+        <EpicTable>
+          <EpicRow header>
+            <EpicHeader width={300} height={44}>
+              First name
+            </EpicHeader>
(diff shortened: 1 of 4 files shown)


jvhoven:
```
/home/travis/build/42BV/ui/src/table/EpicTable/EpicTable.tsx
  12:27  error  'EpicTableLayout' is defined but never used  @typescript-eslint/no-unused-vars
✖ 1 problem (1 error, 0 warnings)
```

Issue #275, opened 2020-01-17
Add ColorPicker component


b-chiarini:
Please add a color picker component. It should display the current color and let you choose the color in a popup window.

PR #276, opened 2020-01-17
feat: added `ColorPicker`.

MrHus:
The color picker is based on `react-color`'s `SketchPicker` surrounded
by a `reactstrap`'s `UncontrolledPopover` with a select and cancel
button.

Also upgraded reactstrap dependency to the latest version.

Refactored the `IconPicker` so it does not need custom CSS, made the
"Clear" button bigger to do this.

Closes: #275


MrHus:
> My two cents for your consideration:
> 
> **Make use of `onChange` instead of `onChangeComplete`**
> 
> `onChangeComplete` makes the color picker UI feel quite unresponsive, as from software like Adobe Photoshop and Sketch I'm used to be able to drag the selection for color and watch the small increments to see if the color I'm picking is the one I like. Although something can be said for the performance implications but, in my opinion, it's worth it.
> 
> **Enable clicking on the selected color to open the color picker UI, as well as the button.**
> 
> Small change that makes usability a bit more friendly.
> 
> **Add `@types/react-color`**.
> 
> More TypeScript is more better! Err... let's keep it at better.

Good catch! I was wondering why the color picker was not as responsive as the examples 👍

Issue #284, opened 2020-01-29
v2.9.0 misses typescript definitions


MrHus:
Rollup complains of a circular dependency maybe that is the problem.

PR #285, opened 2020-01-29
fix: build missed typescript `d.ts` files.

MrHus:
Was somehow caused by moving the `test` directory outside
of the `src` directory. Reverted that change and added
a check in the build step so the existence of the dist/index.d.ts
is made.

Closes: #284


jvhoven:
- [ ] Add ‘src/test’ to ignored files for coverage.

PR #324, opened 2020-02-21
feat: added Tooltip

TimQuist:
Added a Bootstrap style Tooltip component based on Tippy.js

The Tooltip can be wrapped around the target component. On hover the Tooltip will be shown on this component.

Internally the Tooltip wraps the target inside of a CustomTag (this can be a div or a span).
This CustomTag gets a tabIndex={0} for the sake of handeling the forwardRef.

You can add content to the Tooltip which is displayed on hover over the Target component.
This content can be other React components as well.

Styling can be done through several properties such as placement, distance from target, interactivity, maxWidth, bootstrap classNames and css style.

### DIFF
--- a/src/core/Tooltip/Tooltip.tsx
+++ b/src/core/Tooltip/Tooltip.tsx
@@ -1,0 +1,95 @@
+import React, { CSSProperties } from 'react';
+import * as Popper from 'popper.js';
+import Tippy from '@tippy.js/react';
+
+/*** Tooltip component based on the Tippy Library ***/
+interface Props {
+  /**
+   * Target component that, when hovered, will trigger the tooltip to show up.
+   * The target(children) of the tooltip are wrapped into a div.
+   * This is a bypass to not have to forward the ref to the DOM node.
+   */
+  children: React.ReactNode;
+
+  /**
+   * Content shown inside of the tooltip.
+   */
+  content: React.ReactNode;
+
+  /**
+   * Optional alignment relative to the target where the tooltip will be shown.
+   */
+  placement?: Popper.Placement;
+
+  /**
+   * Optional distance that the tooltip will show up relative from the target.
+   * Possible values: number (px), string (with units "rem" only).
+   */
+  distance?: number | string;
+
+  /**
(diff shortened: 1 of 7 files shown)


src/core/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { CSSProperties } from 'react';
+import * as Popper from 'popper.js';
+import Tippy from '@tippy.js/react';
+
+/*** Tooltip component based on the Tippy Library ***/
+interface Props {
+  /** 
+   *Target component that, when hovered, will trigger the tooltip to show up.
```
MrHus: Missing * chars here at the start of the line

src/core/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,94 @@
...
+  interactive?: boolean;
+
+  /**
+   * Optional that allows you to override the default max width of the tooltip
+   * Possible values: number (px), string (with units "rem") or string 'none'.
+   */
+  maxWidth?: number | string;
+
+  /**
+   * Optional that allows you to override the default element that the children get put inside of.
```
MrHus: Document what the default is `span` or `div`.

src/core/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,94 @@
...
+  tag = 'span',
+  className,
+  style
+}: Props) {
+  const Tag = tag;
+
+  return (
+    <Tippy
+      className="border-0"
+      content={<> {content} </>}
```
MrHus: Is the extra fragment needed?

TimQuist: Without the fragments a TS error will occur.

In order to ensure that a string can be given directly to the `Tooltip` content (since this is what will be given to it most of the time) the content of `Tooltip` needs to be `React.ReactNode`. This way both strings and React components can be given to the `Tooltip` `content` directly without wrapping it in fragments.

Since the Tippy `content` expects a React.ReactElement it needs to be wrapped within elements here.

PR #338, opened 2020-02-27
Fix: Defined size as property of SearchInput

### LINKED ISSUE #334: Size is not type-safe in SearchInput props
https://github.com/42BV/ui/blob/abed1e9b89e017756eb1d6892090c24e0d36525c/src/core/SearchInput/SearchInput.tsx#L155

Size is used when showIcon is true, but size is not defined in props. Therefor size is not type-safe.

gidomanders:
Size is used when showIcon is true, but size is not defined in props.
Therefor size is not type-safe. People might not know they could change
the size of the icon or what the available sizes are.
Separated Props into 2 different interfaces, one with showIcon true
which has an optional size, the other with optional showIcon false
where size is not allowed.

Closes #334

### DIFF
--- a/src/core/SearchInput/SearchInput.test.tsx
+++ b/src/core/SearchInput/SearchInput.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow, mount } from 'enzyme';
+import { mount, shallow } from 'enzyme';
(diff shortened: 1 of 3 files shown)


gidomanders:
@MrHus I'll merge the commits and rebase once you've reviewed the new changes.

PR #353, opened 2020-03-02
Fix: Type integrity

### LINKED ISSUE #335: Bug: Type integrity
In Addon, you can specify icon and text without warning. If you add `text?: never` to IconProps, you'll get a warning that icon and text cannot be used together.

gidomanders:
In Addon, you can specify icon and text without warning.
Added `text?: never` to IconProps and `icon?: never` to TextProps so
you'll get a warning that icon and text cannot be used together.
Did the same with ConfirmButton where icon and children should never be
`undefined`, they either have a valid value or they're not specified at
all.

Closes #335

### DIFF
--- a/src/core/ConfirmButton/ConfirmButton.tsx
+++ b/src/core/ConfirmButton/ConfirmButton.tsx
@@ -1,7 +1,13 @@
 import React, { useState } from 'react';
-import { Modal, ModalHeader, ModalBody, ModalFooter } from 'reactstrap';
+import { Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
 
-import Button from '../Button/Button';
+import Button, {
+  BaseProps as ButtonBaseProps,
+  isWithIcon,
+  isWithIconAndText,
+  isWithText,
+  Props as ButtonProps
+} from '../Button/Button';
 import { Color } from '../types';
 import IconType from '../Icon/types';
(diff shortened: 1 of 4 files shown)


src/core/ConfirmButton/ConfirmButton.tsx
```diff
@@ -131,19 +133,34 @@ export default function ConfirmButton({
     onConfirm(event);
   }
 
+  function getButton() {
+    const buttonProps = {
+      onClick: openModal,
+      color,
+      inProgress: !!inProgress
```
MrHus: Can't you just add `children` and `icon` to button props somehow. This way you don't need any
if statement in getButton.

gidomanders: Ignoring the type-safety error with icon and children in Props, I could remove the entire function. Maybe not the ideal situation as I have to ts-ignore, which could cause other problems in the future, but I can't think of another way to solve this without if statements.

src/core/ConfirmButton/ConfirmButton.tsx
```diff
@@ -131,19 +133,23 @@ export default function ConfirmButton({
     onConfirm(event);
   }
 
+  // @ts-ignore
```
MrHus: Can you fix this using my solution I've chatted you.

PR #354, opened 2020-03-02
Fix: Bootstrap sizes should be a registered type

### LINKED ISSUE #339: Bootstrap sizes should be a registered type
Bootstrap sizes like `'xs' | 'sm' | 'md' | 'lg'` are used and defined in multiple files. This should be separated into a BootstrapSize type defined in one place.

gidomanders:
Bootstrap sizes like 'xs' | 'sm' | 'md' | 'lg' are used and defined in
multiple files. This should be separated into a BootstrapSize type
defined in one place.

Closes #339

### DIFF
--- a/src/core/Avatar/Avatar.test.tsx
+++ b/src/core/Avatar/Avatar.test.tsx
@@ -2,12 +2,13 @@ import React from 'react';
 import { shallow, ShallowWrapper } from 'enzyme';
 import toJson from 'enzyme-to-json';
 
-import Avatar, { Size } from './Avatar';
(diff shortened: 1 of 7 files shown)


src/utilities/size/size.ts
```diff
@@ -0,0 +1 @@
+export type BootstrapSize = 'xs' | 'sm' | 'md' | 'lg';
```
MrHus: Belongs in core/types.ts

MrHus:
The SubmitButton and Button have `size` props as well.

PR #366, opened 2020-03-11
fix: ModalPickerSingle should not overflow a fixed-width container

### LINKED ISSUE #365: ModalPickerSingle should not overflow a fixed-width container
When ModalPickerSingle is used within a fixed-width container like a table cell, a long label should not cause the button to be (partly) hidden.

gidomanders:
When ModalPickerSingle is used within a fixed-width container like a
table cell, a long label should not cause the button to be (partly)
hidden.

Closes #365

### DIFF
--- a/src/form/ModalPicker/ModalPickerOpener/ModalPickerOpener.test.tsx
+++ b/src/form/ModalPicker/ModalPickerOpener/ModalPickerOpener.test.tsx
@@ -1,0 +1,75 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { ModalPickerOpener } from './ModalPickerOpener';
+import { adminUser } from '../../../test/fixtures';
+import toJson from 'enzyme-to-json';
+import * as useComponentOverflow from './useComponentOverflow/useComponentOverflow';
+
+describe('Component: ModalPickerOpener', () => {
+  function setup({
+    hasValues = false,
+    withTooltip = false
+  }: {
+    hasValues?: boolean;
+    withTooltip?: boolean;
+  }) {
+    const openModalSpy = jest.fn();
+    const values = hasValues ? adminUser.email : undefined;
+
+    jest
+      .spyOn(useComponentOverflow, 'useComponentOverflow')
+      .mockImplementation(() => withTooltip);
+
+    const modalPickerOpener = shallow(
+      <ModalPickerOpener
+        openModal={openModalSpy}
+        label="Best friend"
+        values={values}
+      />
+    );
+
+    return { modalPickerOpener, openModalSpy };
+  }
+
+  describe('ui', () => {
+    test('without values', () => {
+      const { modalPickerOpener } = setup({});
+      expect(toJson(modalPickerOpener)).toMatchSnapshot(
+        'Component: ModalPickerOpener => ui => without values'
+      );
+    });
+
+    test('with values', () => {
+      const { modalPickerOpener } = setup({ hasValues: true });
+      expect(toJson(modalPickerOpener)).toMatchSnapshot(
+        'Component: ModalPickerOpener => ui => with values'
+      );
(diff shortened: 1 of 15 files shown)


src/form/ModalPicker/ModalPickerOpener/useComponentOverflow/useComponentOverflow.test.ts
```diff
@@ -0,0 +1,95 @@
+import * as useComponentOverflow from './useComponentOverflow';
+
+describe('HoC: useComponentOverflow', () => {
+  test('isOverflowing', () => {
+    expect(
+      useComponentOverflow.isOverflowing({
+        clientWidth: 50,
+        scrollWidth: 50
+      } as HTMLElement)
+    ).toBeFalsy();
```
MrHus: Don't use `toBeFalsy` or `toBeTruthy` just use toBe or toEqual and check the values directly

src/form/ModalPicker/ModalPickerOpener/ModalPickerOpener.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const isOverflowing = useComponentOverflow(ref, values);
+
+  return (
+    <div className="d-flex align-items-center">
+      {values ? (
+        <span className="text-truncate mr-1 position-relative" ref={ref}>
+          {isOverflowing ? (
+            <Tooltip
+              content={values}
+              className="w-100 h-100 position-absolute truncator-tooltip"
```
MrHus: What does the CSS class 'truncator-tooltip' do?

src/form/ModalPicker/ModalPickerOpener/ModalPickerOpener.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { useRef } from 'react';
```
MrHus: Looks strange for the modal picker multiple: 

Screenshot 2020-03-18 at 13 09 54

src/form/ModalPicker/ModalPickerOpener/ModalPickerOpener.test.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import * as ModalPickerOpener from './ModalPickerOpener';
```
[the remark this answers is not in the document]
gidomanders: This was a leftover from when I had to test a function inside the file, which was moved to a HoC later.

src/form/ModalPicker/ModalPickerOpener/ModalPickerOpener.tsx
gidomanders: Also a leftover from development, sorry for not checking for things like that

src/form/ModalPicker/ModalPickerOpener/ModalPickerOpener.tsx
gidomanders: Would you separate values and children to display children in the tooltip?

src/form/ModalPicker/single/ModalPickerSingle.test.tsx
```diff
@@ -475,25 +480,35 @@ describe('Component: ModalPickerSingle', () => {
...
+      ).toBeUndefined();
 
       modalPickerSingle.setProps({
         value: adminUser
       });
 
-      const value = modalPickerSingle.find('span').text();
-      expect(value).toBe('admin@42.nl');
+      expect(
+        modalPickerSingle.find('ModalPickerOpener').props().values
```
MrHus: Snapshots are a weapon for testing html, but it is a very crude tool. When possible use explicit value checks instead. Snapshots are a little too easy to approve.

PR #372, opened 2020-03-16
improvement: withJarb support accepting errors as React.ReactNode

### LINKED ISSUE #369: withJarb support accepting errors a React.ReactNode
This way the user can render errors in a custom for example as a list item.

gidomanders:
This way, the user can render errors in a custom element, for example as
a list item.
Replaced string with ReactNode as type of MetaError to accept custom
nodes as well.
Uses the index as key when mapping ReactNode errors.

Closes #369

### DIFF
--- a/src/form/FormError/FormError.tsx
+++ b/src/form/FormError/FormError.tsx
@@ -3,7 +3,7 @@ import { FormFeedback } from 'reactstrap';
 import { useErrorsForValidator } from '@42.nl/react-error-store';
 
 import { Meta, MetaError } from '../types';
-import { errorMessage, keyForError } from './utils';
+import { errorMessage } from './utils';
 import { useSettledErrors } from './useSettledErrors';
 import { useOnChange } from './useOnChange';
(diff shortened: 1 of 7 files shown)


src/form/FormError/FormError.tsx
```diff
@@ -3,7 +3,7 @@ import { FormFeedback } from 'reactstrap';
 import { useErrorsForValidator } from '@42.nl/react-error-store';
```
MrHus: Good work, but we need a nice example in the stories.

Alter the FormExample story's `firstNameAvailable`:

```
async function firstNameAvailable(value?: string) {
  if (!value) {
    return undefined;
  }

  await sleep(random(2000, 5000));

  return ['Maarten', 'Jeffrey', 'Gido'].includes(value) ? (
    undefined
  ) : (
    <ul>
      <li>First name not available</li>
      <li>You can use Jeffrey</li>
      <li>You can use Maarten</li>
    </ul>
  );
}
```
Maybe change the last name as well

PR #373, opened 2020-03-17
fix: duplicate ID in asyncPage stories

gidomanders:
The console gives a warning about a duplicate ID being used for stories
in AsyncPage. Also, the duplicate ID caused one of the stories not to be
displayed in the menu.
Changed the ID for one of the stories so there is no warning anymore and
both of the stories are displayed in the menu.

### DIFF
--- a/src/core/AsyncPage/AsyncPage.stories.tsx
+++ b/src/core/AsyncPage/AsyncPage.stories.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { useAsync } from 'react-async';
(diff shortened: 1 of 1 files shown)


src/core/AsyncPage/AsyncPage.stories.tsx
```diff
@@ -191,7 +191,7 @@ storiesOf('core|async/AsyncPage', module)
     );
   })
 
-  .add('when empty with custom title', () => {
+  .add('when empty with custom content', () => {
```
MrHus: Can you check asyncList stories as well?

gidomanders: There's no duplicate in that one

PR #374, opened 2020-03-17
fix: code inspection issues

gidomanders:
Removed redundant unit of measure.
Removed redundant local variables.
Removed explicit type argument any.
Renamed birthdate to date of birth.
Changed invalid CSS property value (font-display used where font-style
was meant to be used).
Changed all words with a typo (messed up character order or missing
character).
Closed tags with empty body correctly.
Simplified expressions.
Fixed parameter described in jsdoc that does not appear in function
signature.
Fixed value assigned is never used warning.
Ignored missing await for async function calls.
Ignored variable might not have been initialized warning.
Ignored const reassignment warning.

### DIFF
--- a/src/core/Button/Button.test.tsx
+++ b/src/core/Button/Button.test.tsx
@@ -201,7 +201,7 @@ describe('Component: Button', () => {
 
   describe('events', () => {
     let button: ShallowWrapper;
-    let onClickSpy: jest.Mock<any, any>;
+    let onClickSpy: jest.Mock;
 
     describe('onClick', () => {
       function setup({
--- a/src/form/ImageUpload/ImageUpload.test.tsx
+++ b/src/form/ImageUpload/ImageUpload.test.tsx
@@ -16,8 +16,8 @@ import * as testUtils from '../../test/utils';
 describe('Component: ImageUpload', () => {
   let imgUpload: ShallowWrapper;
 
-  let onChangeSpy: jest.Mock<any, any>;
-  let onBlurSpy: jest.Mock<any, any>;
+  let onChangeSpy: jest.Mock;
+  let onBlurSpy: jest.Mock;
 
   function setup({
     value,
--- a/src/form/ModalPicker/single/ModalPickerSingle.test.tsx
+++ b/src/form/ModalPicker/single/ModalPickerSingle.test.tsx
@@ -19,11 +19,11 @@ import { RenderOptionsOption } from '../../option';
 describe('Component: ModalPickerSingle', () => {
   let modalPickerSingle: ShallowWrapper;
 
-  let onChangeSpy: jest.Mock<any, any>;
-  let onBlurSpy: jest.Mock<any, any>;
-  let fetchOptionsSpy: jest.Mock<any, any>;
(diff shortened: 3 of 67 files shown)


src/form/ValuePicker/ValuePicker.tsx
```diff
@@ -220,7 +221,7 @@ export default function ValuePicker<T>(props: Props<T>) {
   // Have to do `props.multiple === true` here otherwise it does not
```
MrHus: If this indeed works the comment above can be removed

src/core/Button/Button.test.tsx
```diff
@@ -201,7 +201,7 @@ describe('Component: Button', () => {
 
   describe('events', () => {
     let button: ShallowWrapper;
-    let onClickSpy: jest.Mock<any, any>;
+    let onClickSpy: jest.Mock;
```
MrHus: Good to see that this works now.

src/form/ModalPicker/single/ModalPickerSingle.test.tsx
```diff
@@ -400,6 +401,7 @@ describe('Component: ModalPickerSingle', () => {
         const { promise, reject } = testUtils.rejectablePromise();
         addButtonCallbackSpy.mockReturnValueOnce(promise);
 
+        // noinspection ES6MissingAwait
```
MrHus: What does the noinspection do?

src/form/ImageUpload/utils.ts
```diff
@@ -60,7 +60,6 @@ export function cropToAvatarEditorConfig(crop: Crop): AvatarEditorConfig {
 export function calculateScale(scale: number, delta: number): number {
   const zoomFactor = Math.max(Math.abs(delta) / 1000, 0.1);
   const nextScale = delta > 0 ? scale - zoomFactor : scale + zoomFactor;
-  const clamped = Math.min(10, Math.max(1, nextScale));
 
-  return clamped;
+  return Math.min(10, Math.max(1, nextScale));
```
MrHus: I don't like this change makes it less clear that this is a clamp...

src/form/ModalPicker/single/ModalPickerSingle.test.tsx
gidomanders: It is for IntelliJ (and possibly other IDEs) to ignore the warning that there's no `await` before the async function call. Not really a coding problem, but that 'issue' causes IntelliJ to warn you on every commit when you selected the inspect code before commit option. The feature is very good to make sure your code is not going to fail, but it bothers me if I have to check what issues it found and only those come up.

MrHus: Please remove them, I don't want to force people to use Intellij.

src/form/ImageUpload/ImageUpload.test.tsx
```diff
@@ -68,6 +68,7 @@ describe('Component: ImageUpload', () => {
       // @ts-ignore
       const imgUpload = new ImageUpload();
 
+      // noinspection JSConstantReassignment
```
MrHus: Remove all noinspections

PR #375, opened 2020-03-19
improvement: form in EpicTable

### LINKED ISSUE #357: Form in EpicTable
In some cases you might want to display an EpicTable with form fields to easily change a bunch of data in one batch.

### LINKED ISSUE #400: AutoSave forms
It would be nice to have an option to submit forms automatically when a value changes.

gidomanders:
In some cases you might want to display an EpicTable with form fields
to easily change a bunch of data in one batch.
Added EpicForm component which wraps around final-form Form component.
Includes possibility to automatically submit on change.
Added FormButton to submit or reset a Form externally.

Closes #357
Closes #400

### DIFF
--- a/src/form/AutoSave/AutoSave.test.tsx
+++ b/src/form/AutoSave/AutoSave.test.tsx
@@ -1,0 +1,146 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
+
+import { AutoSave } from './AutoSave';
+import * as Active from './useActive';
+import * as PendingPromise from './usePromise';
+import { resolvablePromise } from '../../test/utils';
+
+describe('Component: AutoSave', () => {
+  function setup({
+    activeField,
+    initialValues
+  }: {
+    activeField?: string;
+    initialValues?: { [id: string]: string };
+  }) {
+    const resolvable = resolvablePromise();
+    const onSaveSpy = jest.fn(() => resolvable.promise);
+    const setActiveSpy = jest.fn();
+    jest
+      .spyOn(Active, 'useActive')
+      .mockImplementation(() => [activeField, setActiveSpy]);
+
+    // @ts-ignore
+    const autoSave = shallow(
+      <AutoSave onSave={onSaveSpy} initialValues={initialValues} />
+    );
+
+    return { autoSave, onSaveSpy, setActiveSpy };
+  }
+
+  test('ui', () => {
+    const autoSave = shallow(<AutoSave onSave={jest.fn()} />);
+    expect(toJson(autoSave)).toMatchSnapshot();
+  });
+
+  describe('events', () => {
+    it('should not save on first time field active', () => {
+      const { autoSave, onSaveSpy, setActiveSpy } = setup({});
+
+      // @ts-ignore
+      autoSave
+        .find('FormSpy')
+        .props()
+        // @ts-ignore
+        .onChange({ active: 'test' });
+
+      expect(setActiveSpy).toBeCalledTimes(1);
+      expect(setActiveSpy).toBeCalledWith('test');
+
+      expect(onSaveSpy).toBeCalledTimes(0);
+    });
+
+    it('should not save when same field changes', () => {
+      const { autoSave, onSaveSpy, setActiveSpy } = setup({
+        activeField: 'test'
+      });
+
+      // @ts-ignore
+      autoSave
+        .find('FormSpy')
+        .props()
+        // @ts-ignore
+        .onChange({ active: 'test' });
+
+      expect(setActiveSpy).toBeCalledTimes(0);
+      expect(onSaveSpy).toBeCalledTimes(0);
+    });
+
+    it('should not save when values do not change', () => {
+      const { autoSave, onSaveSpy, setActiveSpy } = setup({
+        activeField: 'test',
+        initialValues: { test: 'test' }
+      });
+
+      // @ts-ignore
+      autoSave
+        .find('FormSpy')
+        .props()
+        // @ts-ignore
+        .onChange({ active: 'friends', values: { test: 'test' } });
+
+      expect(setActiveSpy).toBeCalledTimes(1);
+      expect(setActiveSpy).toBeCalledWith('friends');
+
+      expect(onSaveSpy).toBeCalledTimes(0);
+    });
+
+    it('should wait until saving again until the previous save has finished', async done => {
+      const { promise, resolve } = resolvablePromise();
+      // @ts-ignore
+      jest
+        .spyOn(PendingPromise, 'usePromise')
+        .mockReturnValue({ current: promise });
+
+      const { autoSave, onSaveSpy, setActiveSpy } = setup({
+        activeField: 'test',
+        initialValues: { test: 'test' }
+      });
+
+      // @ts-ignore
+      autoSave
+        .find('FormSpy')
+        .props()
+        // @ts-ignore
+        .onChange({ active: 'friends', values: { test: 'testing' } });
+
+      expect(setActiveSpy).toBeCalledTimes(1);
+      expect(setActiveSpy).toBeCalledWith('friends');
+
+      expect(onSaveSpy).toBeCalledTimes(0);
+
+      try {
+        await resolve();
+
+        expect(onSaveSpy).toBeCalledTimes(1);
+        expect(onSaveSpy).toBeCalledWith({ test: 'testing' });
+
+        done();
+      } catch (e) {
+        console.error(e);
+        done.fail();
+      }
+    });
+
+    it('should save when another field changes', () => {
+      const { autoSave, onSaveSpy, setActiveSpy } = setup({
+        activeField: 'test'
+      });
+
+      // @ts-ignore
+      autoSave
+        .find('FormSpy')
+        .props()
+        // @ts-ignore
+        .onChange({ active: 'friends', values: { test: 'test' } });
+
+      expect(setActiveSpy).toBeCalledTimes(1);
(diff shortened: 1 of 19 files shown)


src/table/EpicTable/cells/EpicForm/EpicForm.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
+
+import { EpicForm } from './EpicForm';
+
+describe('Component: EpicForm', () => {
+  test('ui', () => {
+    const vertical = shallow(
```
MrHus: `const epicForm`

src/table/EpicTable/cells/EpicFormCell/EpicFormCell.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
+
+import { EpicFormCell } from './EpicFormCell';
+
+describe('Component: EpicFormCell', () => {
+  test('ui', () => {
+    const vertical = shallow(
```
MrHus: `epicFormCell`

src/table/FormTable/FormTable.stories.tsx
```diff
@@ -0,0 +1,1091 @@
...
+import { JarbModalPickerSingle } from '../../form/ModalPicker/single/ModalPickerSingle';
+import { pageOf } from '../../utilities/page/page';
+import _ from 'lodash';
+import { Constraints, setConstraints } from '@42.nl/jarb-final-form/lib';
+import { EpicForm } from '../EpicTable/cells/EpicForm/EpicForm';
+import { EpicCell } from '../..';
+import { EpicFormCell } from '../EpicTable/cells/EpicFormCell/EpicFormCell';
+import Button from '../../core/Button/Button';
+
+const constraints: Constraints = {
```
MrHus: This should be configured globally so all examples can use these. Perhaps preview.js in the storybook folder.

src/table/FormTable/FormTable.stories.tsx
```diff
@@ -0,0 +1,1091 @@
...
+              Birth date
+            </EpicHeader>
+            <EpicHeader width={200} height={44}>
+              Sex
+            </EpicHeader>
+          </EpicRow>
+          {persons.map(person => (
+            <EpicRow key={person.id}>
+              <EpicCell width={100} height={52}>
+                {person.id}
```
MrHus: This `id` should ideally contain a submit button for the form. Can a reset button to reset it to the original state. When the user saves it should show a loading indicator until the save is done.

This is not going to be easy...

gidomanders: Do you want a reset + submit in each row?

src/table/FormTable/FormTable.stories.tsx
```diff
@@ -0,0 +1,958 @@
...
+import { pageOf } from '../../utilities/page/page';
+import _ from 'lodash';
+import { setConstraints } from '@42.nl/jarb-final-form/lib';
+import { EpicForm } from '../EpicTable/cells/EpicForm/EpicForm';
+import { EpicCell } from '../EpicTable/cells/EpicCell/EpicCell';
+import { EpicFormCell } from '../EpicTable/cells/EpicFormCell/EpicFormCell';
+import Button from '../../core/Button/Button';
+
+// @ts-ignore
+setConstraints(window.constraints);
```
MrHus: Can this go to preview as well?

src/table/FormTable/FormTable.stories.tsx
MrHus: Actually would be nice to have both...

src/table/FormTable/FormTable.stories.tsx
```diff
@@ -0,0 +1,1108 @@
...
+      height: 0,
+      weight: 0,
+      jobTitle: '',
+      favoriteMovie: undefined,
+      favoriteFood: '',
+      birthDate: undefined,
+      sex: ''
+    });
+    const [personsList, setPersonsList] = useState(
+      _.concat(_.clone(persons), [newPerson])
```
MrHus: Use array destructuring instead: [...persons, newPerson]

src/table/FormTable/FormTable.stories.tsx
```diff
@@ -0,0 +1,1108 @@
...
+      });
+    }
+
+    async function onSubmit(data: Person) {
+      setLoading(data);
+      await sleep(random(2000, 5000));
+      if (data.id === newPerson.id) {
+        setPersonsList(
+          _.concat(_.slice(personsList, 0, personsList.length - 1), [
+            Object.assign(data, { id: Math.random() }),
```
MrHus: You can use destructring instead of Object.assign to copy: {...data id: Math.random() }

src/table/FormTable/FormTable.stories.tsx
```diff
@@ -0,0 +1,1108 @@
...
+        setTimeout(resolve, ms);
+      });
+    }
+
+    async function onSubmit(data: Person) {
+      setLoading(data);
+      await sleep(random(2000, 5000));
+      if (data.id === newPerson.id) {
+        setPersonsList(
+          _.concat(_.slice(personsList, 0, personsList.length - 1), [
```
MrHus: You can use `shift` to remove the first element, and `pop` to remove the last element from an array.

There is also an `unshift` to add an item at the first position of an array, and `push` to add items to the last position the array.

src/table/FormTable/FormTable.stories.tsx
```diff
@@ -0,0 +1,1108 @@
+import React, { useState } from 'react';
+import { storiesOf } from '@storybook/react';
+import { Card } from 'reactstrap';
+import _, { random } from 'lodash';
```
MrHus: Don't import _ from lodash... instead import everything separately.

src/table/EpicTable/cells/EpicForm/EpicForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+    onChange,
+    ...rest
+  } = props;
+  const { odd } = pick(rest, ['odd']) as InjectedProps;
+  const formProps = omit(rest, ['odd']) as FormProps;
+  const [isFormValid, setIsFormValid] = useIsFormValid();
+  const [isFormPristine, setIsFormPristine] = useIsFormPristine();
+  const formApi = useFormApi<FormValues>();
+
+  const debounceCallback = useDebounceCallback(debounce, () => {
```
MrHus, line 79: use `useCallback` to prevent the callback from changing on line 78:

```tsx
 const callback = useCallback(
    () => {
      if (formApi.current && submitOnChange) {
        formApi.current.submit();
      }
    },
    [formApi, submitOnChange],
  );

  const debounceCallback = useDebounceCallback(debounce, callback);
```

src/table/FormTable/FormTable.stories.tsx
```diff
@@ -0,0 +1,1583 @@
...
+          <Pagination
+            className="my-3"
+            page={pageOfPersons}
+            onChange={setPage}
+          />
+        </div>
+      </Card>
+    );
+  })
+  .add('instant edit', () => {
```
MrHus, line 852: Does not work quite right when editing the field when a save is in progress. Steps to reproduce:

1. Change the name of a person.
2. Wait for the debounce to occur and "Processing..." to show up
3. Edit the name again during "Processing..."
4. You will see the old name.

Found this example of how to debounce save in react-final-form: https://codesandbox.io/s/5w4yrpyo7k

Perhaps we can steal from it using `lodash.isEqual` instead of `object-diff`.

src/form/AutoSave/AutoSave.tsx
```diff
@@ -0,0 +1,48 @@
...
+      return;
+    }
+
+    setActive(formState.active);
+
+    if (promise.current) {
+      await promise.current;
+    }
+
+    if (!isEqual(formState.values, values)) {
```
MrHus, line 36: The form submits when nothing changes sometimes because `values` is undefined the first time the form is hit.

I've done an experiment and checking if it is equal to the initial values seems to work:

```tsx
if (isEqual(formState.values, initialValues)) {
   return;
}
```
So I think we should add the initialValues as a prop to the AutoSave.

src/form/AutoSave/AutoSave.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+        .onChange({ active: 'friends', values: { test: 'test' } });
+
+      expect(setActiveSpy).toBeCalledTimes(1);
+      expect(setActiveSpy).toBeCalledWith('friends');
+
+      expect(onSaveSpy).toBeCalledTimes(0);
+      expect(setValuesSpy).toBeCalledTimes(0);
+    });
+
+    it('should await pending promise', async done => {
```
MrHus, line 95: This description is to technical

`should wait until saving again until the previous save has finished`

src/form/AutoSave/AutoSave.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const [active, setActive] = useActive();
+  const [values, setValues] = useValues<FormValues>();
+  const promise = usePromise<T>();
+
+  async function onChange(formState: FormState<FormValues>) {
+    if (active === formState.active) {
+      return;
+    }
+
+    if (!active) {
```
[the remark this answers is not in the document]
gidomanders: But I need to check on the old value to return or not. If the old state was not active and now there is an active field, the active field should be stored and the rest of the function should not be executed. Only if the active field is different than before, the rest of the function should be executed.

PR #411, opened 2020-04-14
improvement: unified form component props

### LINKED ISSUE #65: Form components: unify props
- [x] ModalPickerSingle
- [x] ModalPickerMultiple
- [x] TypeAheadSingle
- [x] TypeAheadMultiple
- [x] Select

Specifically we want to unify the data fetching props and other out of sync properties if possible. It would be nice if changing the name of the component would simply work out of the box.

---

In the plain components (without the `withJarb` construction) sometimes the `name` properties are not defined, which makes them unusable in a `final-form` context `JARB`. 

- [ ] Add `name` properties to form components which do not have them.

gidomanders:
It would be nice if changing the name of the component would simply work
out of the box.

Renamed fetchOptions to options as returning a promise is also allowed
in other components with multiple options (i.e. Select).
Extracted unified props into separate types, extended by all form
component props definitions.

Closes #65


MrHus:
Misssing a couple of new components but otherwise looks good to go, rebase so we can add the missing components.

PR #414, opened 2020-04-16
fix: modal closed when valuepicker total elements narrowed by search

### LINKED ISSUE #412: When searching in a ModalPickerMultiple sometimes the modal closes unexpectedly
When you have a ModalPickerMultiple component with a lot of options and you try to narrow the number of options down by using the search option, the modal closes sometimes.
This seems to happen when you narrow it down too much. 

The cause of this problem lies in the containing ValuePicker component. This component wraps several other components and displays the right one based on the total options. When the ModalPicker narrows the number of options, the value of the total options in the ValuePicker was also changed. The ValuePicker wraps the given optionFetcherCallback in its own callback function. In this function it updates the total number of options. And this callback function is passed to its child components.

gidomanders:
When you have a ModalPickerMultiple component with a lot of options and you try to narrow the number of options down by using the search option, the modal closes sometimes.
This seems to happen when you narrow it down too much.

The cause of this problem lies in the containing ValuePicker component.
This component wraps several other components and displays the right one
based on the total options. When the ModalPicker narrows the number of
options, the value of the total options in the ValuePicker was also
changed. The ValuePicker wraps the given optionFetcherCallback in its
own callback function. In this function it updates the total number of
options. And this callback function is passed to its child components.

Moved setting the total elements to the boot function to prevent total
elements from being changed by a query.

Closes #412

### DIFF
--- a/src/form/ValuePicker/ValuePicker.tsx
+++ b/src/form/ValuePicker/ValuePicker.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { Color } from '../types';
 import { FetchOptionsCallback, OptionEqual, OptionForValue } from '../option';
 import withJarb from '../withJarb/withJarb';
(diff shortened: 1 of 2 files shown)


src/form/ValuePicker/ValuePicker.tsx
```diff
@@ -179,26 +179,16 @@ export default function ValuePicker<T>(props: Props<T>) {
...
-
-      return page;
-    },
-    [fetchOptions]
-  );
-
   useEffect(() => {
     async function boot() {
-      await fetchOptionsCallback('', 1, 1);
+      const page = await fetchOptions('', 1, 1);
```
MrHus, line 184: Does the "switching" between "large" and "small" still work:

`https://42bv.github.io/ui/storybook/?path=/story/form-valuepicker-multiple--basic`

gidomanders: Yes it does :)

PR #419, opened 2020-04-28
feat: input with password strength

### LINKED ISSUE #418: Add input with password strength element
There should be an input element that displays the strength of a password using some kind of progress bar and a list of requirements for the password.

gidomanders:
There should be an input element that displays the strength of a 
password using some kind of progress bar and a list of requirements for 
the password.

Added NewPasswordInput component.

Closes #418

### DIFF
--- a/src/form/NewPasswordInput/NewPasswordInput.stories.tsx
+++ b/src/form/NewPasswordInput/NewPasswordInput.stories.tsx
@@ -1,0 +1,138 @@
+import React, { useState } from 'react';
+import { storiesOf } from '@storybook/react';
+import { FinalForm, Form } from '../story-utils';
+import NewPasswordInput, {
+  isStrongPassword,
+  JarbNewPasswordInput
+} from './NewPasswordInput';
+import { Icon, Tooltip } from '../../index';
+
+storiesOf('Form|NewPasswordInput', module)
+  .add('basic', () => {
+    const [password, setPassword] = useState('');
+
+    return (
+      <Form>
+        <NewPasswordInput
+          id="password"
+          label="Password"
+          placeholder="Please enter your password"
+          value={password}
+          onChange={setPassword}
+        />
+      </Form>
+    );
+  })
+  .add('icon', () => {
+    const [password, setPassword] = useState('');
+
+    return (
+      <Form>
+        <NewPasswordInput
+          id="password"
+          label="Password"
+          placeholder="Please enter your password"
+          value={password}
+          onChange={setPassword}
+          addon={{
+            icon: 'lock',
+            position: 'left'
+          }}
+        />
+      </Form>
+    );
+  })
+  .add('without placeholder', () => {
+    const [password, setPassword] = useState('');
+
+    return (
+      <Form>
+        <NewPasswordInput
+          id="password"
+          label="Password"
+          value={password}
+          onChange={setPassword}
+        />
+      </Form>
+    );
+  })
+  .add('without label', () => {
+    const [password, setPassword] = useState('');
+
+    return (
+      <Form>
+        <NewPasswordInput
+          placeholder="Please enter your password"
+          value={password}
+          onChange={setPassword}
+        />
+      </Form>
+    );
+  })
+  .add('with custom label', () => {
+    const [password, setPassword] = useState('');
+
+    return (
+      <Form>
+        <NewPasswordInput
+          id="password"
+          label={
+            <div className="d-flex justify-content-between">
+              <span>Password</span>
+              <Tooltip
+                className="ml-1"
+                content="Your password should be secret"
+              >
+                <Icon icon="info" />
+              </Tooltip>
+            </div>
+          }
+          placeholder="Please enter your password"
+          value={password}
+          onChange={setPassword}
+        />
+      </Form>
+    );
+  })
+  .add('without meter', () => {
+    const [password, setPassword] = useState('');
+
+    return (
+      <Form>
+        <NewPasswordInput
+          id="password"
+          label="Password"
+          placeholder="Please enter your password"
+          value={password}
+          onChange={setPassword}
+          showMeter={false}
+        />
+      </Form>
+    );
+  })
+  .add('jarb', () => {
+    return (
+      <FinalForm>
+        <JarbNewPasswordInput
+          name="password"
+          jarb={{ validator: 'User.password', label: 'Password' }}
+          id="password"
+          label="password"
+          placeholder="Please enter your password"
+          validators={[
+            isStrongPassword(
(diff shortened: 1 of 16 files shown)


src/form/PasswordStrength/PasswordStrength.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+import Input from '../Input/Input';
+import PasswordStrength from './PasswordStrength';
+
+storiesOf('Form|PasswordStrength', module)
+  .add('basic', () => {
+    const [password, setPassword] = useState('');
+
+    const passwordRules = [
+      {
+        id: 'hasLowerCase',
```
MrHus, line 13: It would be nice if we could define all these rules already so that the user only needs to import them.
The user could then still add custom rules.

```tsx
import { hasUpperCase } from `./PasswordStrength/rules`;

passwordRules = [
   hasLowerCase,
   hasUpperCase
]

File: src/form/PasswordStrength/PasswordStrength.stories.tsx [hunk 2]
```diff
@@ -0,0 +1,162 @@
...
+        compliant: true,
+        label: 'Must not contain any space',
+        regex: '^\\S*$'
+      }
+    ];
+
+    return (
+      <Form>
+        <Input type="password" onChange={setPassword} />
+        <PasswordStrength password={password} initialRules={passwordRules} />
```
MrHus, line 53: Perhaps `rules` is a better name than `initialRules` since they are synced.

src/form/PasswordStrength/useRules/useRules.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react';
+import { Rule } from '../PasswordStrength';
+
+export function useRules(initialRules: Rule[], password: string) {
+  const [rules, setRules] = useState(initialRules);
+
+  useEffect(() => {
+    setRules(
+      initialRules.map(rule => {
```
MrHus, line 9: Should this not be `rules` instead of `initialRules`?

src/form/PasswordStrength/PasswordStrength.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+type Props = {
+  password: string;
+  initialRules: Rule[];
+  showMeter?: boolean;
+  showRulesList?: boolean;
+};
+
+export default function PasswordStrength(props: Props) {
```
MrHus, line 22: Add it to the `index` export otherwise it cannot be imported.

MrHus, line 22: Also add it to the `Example` story so it can be viewed with all other form elements.

src/form/PasswordStrength/PasswordStrength.stories.tsx
```diff
@@ -0,0 +1,162 @@
+import React, { useState } from 'react';
+import { storiesOf } from '@storybook/react';
+import { Form } from '../story-utils';
+import Input from '../Input/Input';
+import PasswordStrength from './PasswordStrength';
+
+storiesOf('Form|PasswordStrength', module)
```
MrHus, line 7: Add common variants:

- Jarb
- label without label
- custom validator

src/form/PasswordStrength/PasswordStrength.stories.tsx
MrHus: We can then also test these rules with a unit test.

src/form/PasswordStrength/PasswordStrength.tsx
```diff
@@ -0,0 +1,66 @@
...
+              ? 'warning'
+              : 'danger'
+          }
+          value={meterWidth}
+          className="mb-2"
+          style={{ height: 5 }}
+        />
+      ) : null}
+      {showRulesList ? (
+        <ListGroup className="mb-2">
```
MrHus, line 50: The list group is a bit overkill, this using only divs looks nice:

Screenshot 2020-04-28 at 11 19 50

src/form/PasswordStrength/PasswordStrength.tsx
```diff
@@ -0,0 +1,66 @@
...
+  initialRules: Rule[];
+  showMeter?: boolean;
+  showRulesList?: boolean;
+};
+
+export default function PasswordStrength(props: Props) {
+  const {
+    password,
+    initialRules,
+    showMeter = true,
```
MrHus, line 26: I don't think the rulesList can ever not be optional. Without rules this component is useless.

src/form/PasswordStrength/PasswordStrength.stories.tsx
gidomanders: I thought it should be in the form directory because it's only used in forms. It is actually not an input though, so it doesn't have a jarb variant etc.

MrHus: Aha I overlooked that fact, but it would be nice for our end users then if the rules would also affect a Jarb valid state, and to show an example of how to do this. Perhaps it should also be a validator?

src/form/NewPasswordInput/NewPasswordInput.tsx
```diff
@@ -0,0 +1,203 @@
...
+      <PasswordStrength
+        password={inputProps.value ?? ''}
+        rules={rules}
+        {...passwordStrengthProps}
+      />
+    </>
+  );
+}
+
+export function JarbNewPasswordInput(
```
MrHus, line 107: Why not use the `withJarb` hoc here?

src/form/NewPasswordInput/NewPasswordInput.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+  })
+  .add('jarb', () => {
+    return (
+      <FinalForm>
+        <JarbNewPasswordInput
+          name="password"
+          jarb={{ validator: 'User.password', label: 'Password' }}
+          id="password"
+          label="password"
+          placeholder="Please enter your password"
```
MrHus, line 125: Add the `isSuperman` custom validator as an example.

src/form/NewPasswordInput/NewPasswordInput.tsx
MrHus: You really need to use the withJarb, because it does more than what your version does not. To make the validations part of the jarb variant. Use the same trick we use elsewhere, which is that the user must explicitly add the validators: see the FileInput's `requireFile` as an example.

Then create a function which returns a validator based on the rules, then the user can use this validator manually when using the jarb variant.

Issue #421, opened 2020-05-01
Popover component


MrHus:
Add custom popover component.

PR #422, opened 2020-05-01
feat: added custom popover component.

MrHus:
The `Popover` component used tippy.js like the `Tooltip` component does.

Updated `@tippy.js/react` from `3.1.1` to `4.0.2`.

Closes: #421


src/core/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+          <Popover target={<Tag color="success" text="default offset" />}>
+            default offset
+          </Popover>
+
+          <Popover
+            target={<Tag color="success" text="very close" />}
+            offset={-100}
+          >
+            very close
```
gidomanders, line 182: I think you mean negative offset or far away to the other side or something like that?

PR #433, opened 2020-06-18
improvement: LoadingPage now accepts custom height and style.

### LINKED ISSUE #431: LoadingPage: optionally allow setting manual height, and allow style properties,
LoadingPage: allow setting manual height, this should remove the default 100% vh

MrHus:
When the new height property is optional. When it is provided the height
of the popover is no longer the fullscreen.

Refactored the hook `LoadingPage` used to its own file, and called if
`useShowAfter`. Also exposing the `useShowAfter` hook now so users of
our library can use it as well.

Closes #431

### DIFF
--- a/src/core/useShowAfter/useShowAfter.tsx
+++ b/src/core/useShowAfter/useShowAfter.tsx
@@ -1,0 +1,26 @@
+import { useEffect, useState } from 'react';
+
+/**
+ * Returns `false` but after a predetermined `after` time returns
+ * `true`.
+ *
+ * Useful for wanting to show content only after a time. For example
+ * to not show a loading spinner to soon, only
+ * after it takes a while.
+ *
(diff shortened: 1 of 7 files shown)


src/core/useShowAfter/useShowAfter.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Returns `false` but after a predetermined `after` time returns
+ * `true`.
+ *
+ * Useful for wanting to show content only after a time. For example
+ * to not show a loading spinner to soon, only
+ * after it takes a while.
+ */
+export function useShowAfter(after: number): boolean {
```
gidomanders, line 11: It's not clear that `after` is the number of milliseconds. Maybe add that to the comments?

PR #435, opened 2020-06-18
improvement: RadioGroup, and the modalpickers can now clear the value.

### LINKED ISSUE #428: ModalPickerSingle / ModalPickerMultiple / RadioGroup clear button
The ModalPickerSingle / RadioGroup should have a clear button.

Should be styled like the clear button from the color picker and icon picker:

https://42bv.github.io/ui/storybook/?path=/story/form-colorpicker--basic

We should create a new component called a `TextButton` so we can extract the common bits in here. Also we should expose the `TextButton` to our users. In order to do this it should become a component.

MrHus:
The clear behavior in the `ModalPickerMultiple` and `ModalPickerSingle`
acts and looks and acts the same as the `IconPicker` and `ColorPicker`.

The `RadioGroup` renders the clear button above the radio inputs. But
only when the new `canClear` props is set to `true`.

The `ValuePicker`'s `RadioGroup` will set the `canClear` props to
`true`. This way all `ValuePicker` variants van clear the selection.

Extracted the shared "clear" button into a `TextButton` component which
is also exposed.

Closes #428

### DIFF
--- a/src/form/RadioGroup/RadioGroup.test.tsx
+++ b/src/form/RadioGroup/RadioGroup.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow, ShallowWrapper } from 'enzyme';
+import { shallow } from 'enzyme';
 import toJson from 'enzyme-to-json';
 
 import RadioGroup, { Text } from './RadioGroup';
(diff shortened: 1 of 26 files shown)


src/form/RadioGroup/RadioGroup.test.tsx
```diff
@@ -42,27 +39,34 @@ describe('Component: RadioGroup', () => {
       onBlur: onBlurSpy,
       error: 'Some error',
       valid: false,
-      horizontal
+      horizontal,
+      hasLabel: hasLabel ? 'Subject' : undefined,
```
gidomanders, line 43: The hasLabel should be label. There is no hasLabel property on the RadioGroup component.

MrHus:
> Where are the stories for ModalPickerMultiple, ModalPickerSingle and ValuePicker?

@gidomanders The 'clear' button will always show for these components. So the existing stories will show the clear button. They look the same as the ColorPicker and IconPicker.

PR #439, opened 2020-06-23
fix: DisplayValues type in ModalPickerSingle

### LINKED ISSUE #438: ModalPicker single and multiple displayValues type is wrong
Currently they both share the following DisplayValues type:

`DisplayValues<T> = (values?: T | T[] | React.ReactNode) => React.ReactNode;`

This is wrong:

ModalPickerSingle should have:

`(values?: T) => React.ReactNode;`

And multiple should have:

`values?: T[]) => React.ReactNode;`

The type should not be shared!

gidomanders:
The ModalPickerSingle doesn't have multiple selected values, so
when specifying displayValues you should not have to deal with a
multi-value type.

Added a DisplayValue type that only deals with a single value.

Closes #438


src/form/ModalPicker/multiple/ModalPickerMultiple.tsx
```diff
@@ -17,10 +17,11 @@ import {
...
   RenderOptionsOption
 } from '../../option';
-import {
-  DisplayValues,
-  ModalPickerOpener
-} from '../ModalPickerOpener/ModalPickerOpener';
+import { ModalPickerOpener } from '../ModalPickerOpener/ModalPickerOpener';
+
+export type DisplayValues<T> = (
+  values?: T[] | React.ReactNode
```
MrHus, line 23: Remove the React.ReactNode.

src/form/ModalPicker/single/ModalPickerSingle.tsx
```diff
@@ -16,10 +16,9 @@ import {
   RenderOptions,
   RenderOptionsOption
 } from '../../option';
-import {
-  DisplayValues,
-  ModalPickerOpener
-} from '../ModalPickerOpener/ModalPickerOpener';
+import { ModalPickerOpener } from '../ModalPickerOpener/ModalPickerOpener';
+
+export type DisplayValue<T> = (values?: T | React.ReactNode) => React.ReactNode;
```
MrHus, line 21: Remove the React.ReactNode.

PR #441, opened 2020-06-24
fix: DateTimeInput does now allow an empty value

b-chiarini:
The check on if the value is invalid now only checks the valid prop or
the hasFormatError indicator.
The hasFormatError is only set to true when the manually entered value
is not a valid date and is not empty.


src/form/DateTimeInput/DateTimeInput.test.tsx
```diff
@@ -205,6 +205,36 @@ describe('Component: DateTimeInput', () => {
...
+        setup({});
+
+        const dateTime = dateTimeInput.find(Datetime);
+        // @ts-ignore
+        dateTime.props().onChange('2000-__-__ __:__:__');
+
+        const updatedDateTime = dateTimeInput.find(Datetime);
+        // @ts-ignore
+        expect(updatedDateTime.props().inputProps.invalid).toBe(true);
+        expect(toJson(dateTimeInput)).toMatchSnapshot(
```
MrHus, line 218: Remove the snapshot here since we do a direct check if it is valid.

src/form/DateTimeInput/DateTimeInput.test.tsx
```diff
@@ -205,6 +205,36 @@ describe('Component: DateTimeInput', () => {
...
+          'Component: DateTimeInput => ui => Invalid DateTimeInput with value'
```
MrHus, line 234: Also remove this snapshot

MrHus:
Please add `Closes #{issue number}` in you commit message.

PR #453, opened 2020-07-17
improvement: OpenCloseModal now has icons and supports a sticky footer

### LINKED ISSUE #452: OpenCloseModal support icons and sticky footer.
The OpenCloseModal does not show icons on the save and cancel buttons.

Also the footer should be sticky so the user can always press the buttons even when the content is
very large.

MrHus:
The sticky footer is enabled by default but can be disabled. When
enabled it will make the footer stay with the user, this way the user
can always press the save and cancel buttons.

The save and cancel buttons also have an icon by default now. The
icons can optionally be changed by the user. The default icons are
`save` and `cancel`.

Also added a story on how to properly use a final-form `Form` inside
of the `OpenCloseModal`. It uses the same form as the `FormExample`
which exports it now. Added it on two locations: with the `FormExample`
and the `OpenCloseModal`, so it can be found more easily.

Closes #452

### DIFF
--- a/src/core/OpenCloseModal/OpenCloseModal.stories.tsx
+++ b/src/core/OpenCloseModal/OpenCloseModal.stories.tsx
@@ -1,11 +1,14 @@
 import React, { useState } from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
+import { Form } from 'react-final-form';
+import { Card } from 'reactstrap';
 
 import { OpenCloseModal } from './OpenCloseModal';
-import { Card } from 'reactstrap';
 import Button from '../Button/Button';
 import RadioGroup from '../../form/RadioGroup/RadioGroup';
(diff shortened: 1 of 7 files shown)


src/core/OpenCloseModal/OpenCloseModal.stories.tsx
```diff
@@ -141,3 +197,56 @@ storiesOf('core|OpenCloseModal', module)
...
+          >
+            <TotalForm />
+          </OpenCloseModal>
+        )}
+      </Form>
+
+      <p className="mt-2">
+        Note: the <pre className="d-inline text-info">Form</pre> element should
+        wrap the <pre className="d-inline text-info">OpenCloseModal</pre>
+        not the other way around. Otherwise you cannot use the inProgress prop.
```
gidomanders, line 243: Also, you cannot use submit in onSave

Issue #447, opened 2020-06-30
ModalPickerMultiple cannot deselect item with custom renderOptions


MrHus:
Seems to me that the itemClicked, should for the filter use the isOptionEqual otherwise the identity will not match.

```tsx
itemClicked(item: T, isChecked: boolean) {
    let selected = this.state.selected as T[];

    if (isChecked) {
      selected = selected.filter(v => v !== item);
    } else {
      selected.push(item);
    }

    this.setState({ selected });
  }

  fetchOptions(query: string) {
    this.setState({ query }, () => {
      this.loadPage(1);
    });
  }
```

MrHus:
It affects the `CheckboxMultipleSelect` as well.

PR #455, opened 2020-07-17
fix: CheckboxMultipleSelect, and ModalPickerMultiple deselect was broken

MrHus:
The deselect broke when using a custom `isOptionEqual` callback. The
idea of the `isOptionEqual` is that you can say that two objects are
equal based on something other than the default reference equality.
The canonical usecase being checking two `id`'s.

In both the `CheckboxMultipleSelect`, and `ModalPickerMultiple` when
clicking an option a routine is executed when an option is selected,
which should remove it when it is already selected.

In both cases de deselection uses a filter:

```ts
   selected = selected.filter(v: T) => v !== item);
```

Of course this filter uses simple reference equality. Which in cases
when using `isOptionEqual` breaks, because the user will provide us with
different `objects` each time, which of course breaks selection.

The fix is using the `isOptionEqual` instead, simple enough.

However how did this bug exist in the first place. We use objects
everywere in our test and stories? The answer is that the fixtures
are frozen objects. This makes reference equality work... by accident.

Changed all fixtures to be functions instead, this way we always create
new instances, and then these bugs are caught more easily. Of course
this meant changing the code which use the fixtures.

Closes: #447

### DIFF
--- a/src/form/CheckboxMultipleSelect/CheckboxMultipleSelect.test.tsx
+++ b/src/form/CheckboxMultipleSelect/CheckboxMultipleSelect.test.tsx
@@ -24,14 +24,18 @@ describe('Component: CheckboxMultipleSelect', () => {
     text,
     hasPlaceholder = true,
     hasLabel = true,
-    horizontal
+    horizontal,
+    hasIsOptionEqual = false,
+    options
   }: {
     value?: User[];
     isOptionEnabled?: OptionEnabledCallback<User>;
+    options?: User[];
     text?: Text;
     hasPlaceholder?: boolean;
     hasLabel?: boolean;
     horizontal?: boolean;
+    hasIsOptionEqual?: boolean;
   }) {
     onChangeSpy = jest.fn();
     onBlurSpy = jest.fn();
@@ -41,12 +45,15 @@ describe('Component: CheckboxMultipleSelect', () => {
       text,
       isOptionEnabled,
       optionForValue: (user: User) => user.email,
-      options: [adminUser, coordinatorUser, userUser],
+      options: options ? options : [adminUser(), coordinatorUser(), userUser()],
       value,
       onChange: onChangeSpy,
       onBlur: onBlurSpy,
       error: 'Some error',
-      horizontal
+      horizontal,
+      isOptionEqual: hasIsOptionEqual
+        ? (a: User, b: User) => a.id === b.id
+        : undefined
     };
 
     if (hasLabel) {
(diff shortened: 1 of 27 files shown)


src/form/CheckboxMultipleSelect/CheckboxMultipleSelect.test.tsx
```diff
@@ -205,64 +212,133 @@ describe('Component: CheckboxMultipleSelect', () => {
...
+        expect(onChangeSpy).toHaveBeenCalledTimes(3);
+        expect(onChangeSpy).toHaveBeenCalledWith([coordinator]);
+
+        // Check that selected is a copy of value
+        expect(onChangeSpy.mock.calls[2][0]).not.toBe(value);
+
+        expect(onBlurSpy).toHaveBeenCalledTimes(3);
+      });
+
+      it('should when there a custom isOptionEqual use that for the equality check', () => {
```
gidomanders, line 280:
```suggestion
      it('should when there is a custom isOptionEqual use that for the equality check', () => {
```

gidomanders: Also shouldn't you use a spy to check if the custom isOptionEqual is actually called?

src/form/ModalPicker/multiple/ModalPickerMultiple.test.tsx
```diff
@@ -422,162 +435,286 @@ describe('Component: ModalPickerMultiple', () => {
...
+          expect(modalPickerMultiple.state().selected).toEqual([
+            admin,
+            coordinator
+          ]);
+        });
 
-      modalPickerMultiple.setState({
-        page: pageOfUsers,
-        selected: [adminUser, userUser]
+        it('should when there a custom isOptionEqual use that for the equality check', () => {
```
gidomanders, line 575:
```suggestion
        it('should when there is a custom isOptionEqual use that for the equality check', () => {
```

gidomanders: Same here, maybe a spy to check if the isOptionEqual function is actually used?

src/form/CheckboxMultipleSelect/CheckboxMultipleSelect.test.tsx
MrHus: No the spy is not needed. We check for the behaviour extensively. This way when the implementation changes te tests do not have to be altered.

PR #463, opened 2020-07-22
fix: options mapping

gidomanders:
Every form element with multiple options use the label as key for
mapping, but the label is not always unique. This sometimes causes more
options to be displayed than actually should be displayed.

Added optional uniqueKeyForValue property to get unique property for an
item.

### DIFF
--- a/src/form/CheckboxMultipleSelect/CheckboxMultipleSelect.tsx
+++ b/src/form/CheckboxMultipleSelect/CheckboxMultipleSelect.tsx
@@ -9,10 +9,12 @@ import Spinner from '../../core/Spinner/Spinner';
 import { Color } from '../types';
 import {
   isOptionSelected,
+  keyForOption,
   OptionEnabledCallback,
   OptionEqual,
   OptionForValue,
-  OptionsFetcher
+  OptionsFetcher,
+  UniqueKeyForValue
 } from '../option';
 import { t } from '../../utilities/translation/translation';
 import { doBlur } from '../utils';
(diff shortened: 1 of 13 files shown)


src/form/option.ts
```diff
@@ -73,15 +80,41 @@ export function isOptionSelected<T>({
...
+  uniqueKeyForValue?: UniqueKeyForValue<T>;
+  optionForValue: OptionForValue<T>;
+};
+
+export function keyForOption<T>({
+  option,
+  uniqueKeyForValue,
+  optionForValue
+}: KeyForOptionConfig<T>) {
+  // @ts-ignore
```
MrHus, line 113: Het lijkt me dat `uniqueKeyForValue` als die defined is die moet winnen.

src/form/option.ts
```diff
@@ -7,6 +7,11 @@ import { ReactNode } from 'react';
  */
 export type OptionForValue<T> = (value: T) => string;
 
+/**
+ * Callback to determine the unique identifier for a given value of type T.
+ */
+export type UniqueKeyForValue<T> = (value: T) => string;
```
MrHus, line 13: Should `UniqueKeyForValue` not be used in all components that use `isOptionSelected`, and not just the modal pickers?

src/form/option.ts
```diff
@@ -73,15 +80,41 @@ export function isOptionSelected<T>({
...
+      const key = keyForOption({ option, uniqueKeyForValue, optionForValue });
+      return (
+        key ===
+        keyForOption({ option: value, uniqueKeyForValue, optionForValue })
+      );
     }
   }
 }
+
+type KeyForOptionConfig<T> = {
```
MrHus, line 102: There are no tests for the new behaviour.

src/form/CheckboxMultipleSelect/CheckboxMultipleSelect.tsx
```diff
@@ -107,6 +109,13 @@ interface BaseProps<T> {
    * Defaults to `false`
    */
   horizontal?: boolean;
+
+  /**
+   * Optionally callback to get unique property for an item.
+   * This is used in mapping the options in the modal.
+   * Defaults to the 'id' property.
```
MrHus, line 116: Defaults to the 'id' property when it exists on T, otherwise it will default to the label.

PR #470, opened 2020-07-30
fix: DateTimeInput displaying empty label

gidomanders:
The DateTimeInput displays a label with only the format while it should
not display any label when the label property is not defined.

Converted MarkedAsRequiredLabel into hook to be able to return label as
is when no required mark should be added.


src/form/withJarb/useMarkedAsRequiredLabel/useMarkedAsRequiredLabel.tsx
```diff
@@ -1,16 +1,19 @@
...
+
 import { isRequired } from '@42.nl/jarb-final-form';
-import { getConfig } from '../../../../config/config';
-import { t } from '../../../../utilities/translation/translation';
 
-type Props = {
+import { getConfig } from '../../../config/config';
+import { t } from '../../../utilities/translation/translation';
+
+export function useMarkedAsRequiredLabel({
```
MrHus, line 8: Why does this have to become a `hook`?

src/form/withJarb/useMarkedAsRequiredLabel/useMarkedAsRequiredLabel.test.tsx
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
+import { setConstraints } from '@42.nl/jarb-final-form';
+import { configure } from '../../../config/config';
+import Tooltip from '../../../core/Tooltip/Tooltip';
+import { useMarkedAsRequiredLabel } from './useMarkedAsRequiredLabel';
+import { renderHook } from '@testing-library/react-hooks';
+
+describe('HoC: useMarkedAsRequiredLabel', () => {
```
MrHus, line 8: This is not a higher order component but a hook. A higher order component is a function which returns a component definition, which can be used as a component.

src/form/withJarb/useMarkedAsRequiredLabel/useMarkedAsRequiredLabel.tsx
```diff
@@ -1,16 +1,19 @@
...
+}: {
   label?: React.ReactNode;
   validator: string;
-};
-
-export function MarkedAsRequiredLabel({ label, validator }: Props) {
+}) {
   if (!label) {
-    return null;
+    return label;
```
MrHus, line 16: If this was not a hook you could return `<>{label}</>` right?

src/form/withJarb/useMarkedAsRequiredLabel/useMarkedAsRequiredLabel.tsx
gidomanders: If it's a component, the label in the DateTimeInput is never empty

src/form/withJarb/useMarkedAsRequiredLabel/useMarkedAsRequiredLabel.tsx
gidomanders: Yes that's how it worked previously, but if it's a component you can never check for an empty label because the label is always a component

PR #482, opened 2020-10-09
fix: ImageUpload replace file extension to match output

### LINKED ISSUE #480: ImageUpload does not convert the extension to .png if the extension of of another image
It currently it accepts has an "image/*"  and converts it to a .png when the done button is clicked.
But the extension remains the same, even through the image is now a .png. 

This is not a problem because the browser ignores the extension, but is would be nice to convert the extension for viewers of the file.

I propose we convert the name of the file and add the new extension and remove the old one. We should have a prop `keepOldFileExtension` which is optional, and `false` by default, but when it is `true` it will not convert the extension.

gidomanders:
ImageUpload accepts any image, but the output is always a PNG file. The
output file doesn't have the .png extension though. This is not a
problem for browsers, as they don't use the extension, but the
extension should match the contents.

Added a function that replaces the file extension with .png.
Added an optional property to keep the original file name.

Closes #480

### DIFF
--- a/src/form/ImageUpload/ImageUpload.tsx
+++ b/src/form/ImageUpload/ImageUpload.tsx
@@ -14,7 +14,8 @@ import {
   calculateScale,
   cropToAvatarEditorConfig,
   dataUrlToFile,
-  getPicaInstance
+  getPicaInstance,
+  replaceFileExtension
 } from './utils';
 
 export interface Text {
@@ -102,6 +103,13 @@ interface BaseProps {
    * Optionally customized text you want to use in this component.
    */
   text?: Text;
+
+  /**
+   * Optionally whether to keep the original file extension or to
+   * replace it with .png (because the output is always a PNG file).
+   * Defaults to false.
+   */
+  keepOriginalFileExtension?: boolean;
 }
 
 interface WithoutLabel extends BaseProps {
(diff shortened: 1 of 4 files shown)


src/form/ImageUpload/ImageUpload.tsx
```diff
@@ -239,15 +246,24 @@ export default class ImageUpload extends Component<Props, State> {
       );
 
       const dataUrl = picaCanvas.toDataURL('image/png', 1.0);
-      const file = dataUrlToFile(dataUrl, this.state.fileName);
+      const fileName = keepOriginalFileExtension
+        ? this.state.fileName
+        : this.replaceExtension(this.state.fileName);
+      console.log(keepOriginalFileExtension, this.state.fileName, fileName);
```
MrHus, line 252: Remove console.log

src/form/ImageUpload/ImageUpload.tsx
```diff
@@ -239,15 +246,24 @@ export default class ImageUpload extends Component<Props, State> {
...
 
       this.props.onChange(file);
       doBlur(this.props.onBlur);
 
-      this.setState({ mode: Mode.FILE_SELECTED, imageSrc: dataUrl });
+      this.setState({ mode: Mode.FILE_SELECTED, imageSrc: dataUrl, fileName });
     }
   }
 
+  replaceExtension(fileName: string) {
```
MrHus, line 262: Make this a util for the ImageUpload in its own file, and aggressively test it.

MrHus (approved):

PR #483, opened 2020-10-15
feat: Input can be made disabled and read-only

arjanvlek:
Added the ability to make form inputs `disabled` or `readOnly`. These
behave exactly as in traditional forms: Disabled inputs are not editable
and not included on submission. Read-only inputs are not editable but
are included on submission


MrHus:
We will not add disabled and readonly inputs due to it being somewhat vague UX.

https://uxdesign.cc/why-you-shouldnt-include-disabled-interaction-elements-in-your-design-system-76a2d4307faf

PR #487, opened 2020-10-26
improvement: Pagination component added next and previous buttons.

### LINKED ISSUE #486: Pagination component add previous and next page buttons
The previous button should be placed on the left of the pagination, the next button on the right.

These buttons will be icons only, use the same icons as on the Pager: https://42bv.github.io/ui/storybook/?path=/docs/core-pager--default.

By default this behaviour will be turned "on" but an prop `showPreviousAndNextButtons` should be added which is optional and defaults to `true`.

The previous button should only be enabled when not on the `first` page. The next button should only be enabled when not on the `last` page.

MrHus:
The `Pagination` will now by default show a next and previous buttons
in the form of an arrow icon. Clicking the buttons will move the the
next or previous page. The previous button is disabled when the user
is already on the first page, the next button is disabled when the
user is already on the last page.

A new property called `showPreviousAndNextButtons` on `Pagination` will
hide the new next and previous button when `showPreviousAndNextButtons`
is `false`.

Changed the `Pagination` components story so it actually updated the
page instead of loggin an action.

Also added documentation on the `pageOf` utility.

Removed nested `Text` prop from the `Pager` this was probably a copy
paste error.

Closes #486


src/core/Pagination/Pagination.test.tsx
```diff
@@ -45,33 +48,59 @@ describe('Component: Pagination', () => {
...
       expect(toJson(pagination)).toMatchSnapshot(
         'Component: Pagination => ui => no next'
       );
     });
+
+    test('with previous and next buttons', () => {
+      setup({ number: 10, totalPages: 10, showPreviousAndNextButtons: true });
+
+      expect(toJson(pagination)).toMatchSnapshot(
+        'Component: Pagination => ui => no next'
```
gidomanders, line 78:
```suggestion
        'Component: Pagination => ui => with previous and next buttons'
```

src/core/Pagination/Pagination.test.tsx
```diff
@@ -45,33 +48,59 @@ describe('Component: Pagination', () => {
...
+    });
+
+    it('should disable the next button when on the last page', () => {
+      setup({ number: 3, totalPages: 3, showPreviousAndNextButtons: true });
+
+      const items = pagination.find('PaginationItem');
+
+      // Next button
+      expect(items.at(4).props().disabled).toBe(true);
+    });
```
gidomanders, line 98: Maybe also test they're not disabled on another page?

src/core/Pagination/Pagination.test.tsx
```diff
@@ -45,33 +48,59 @@ describe('Component: Pagination', () => {
...
   });
 
   describe('events', () => {
     it('should call onChange a pagination links are clicked', () => {
```
gidomanders, line 102:
```suggestion
    it('should call onChange as pagination links are clicked', () => {
```

src/utilities/page/page.ts
```diff
@@ -1,5 +1,13 @@
 import { Page } from '@42.nl/spring-connect';
 
+/**
+ * Creates takes an array of content and turns that array into a
```
gidomanders, line 4:
```suggestion
 * Takes an array of content and turns that array into a
```

gidomanders (changes requested):

PR #488, opened 2020-11-03
feat: Added ConfirmModal to trigger user confirmation without button

arjanvlek:
Added `ConfirmModal`, which is based on `OpenCloseModal` and offers
the user a confirmation before performing an action.

A confirmation dialog was previously only accessible through
`ConfirmButton`. While `ConfirmButton` remains the primary choice to
perform operations that need confirmation from the user, in some cases
it could be needed to offer a confirmation modal through other means.

An example is a `ButtonDropdown` which contains a remove action that
needs confirmation. In such case, we want to use the generic confirm
modal but don't want to render a button.

### DIFF
--- a/src/core/ConfirmModal/ConfirmModal.stories.tsx
+++ b/src/core/ConfirmModal/ConfirmModal.stories.tsx
@@ -1,0 +1,96 @@
+import { storiesOf } from '@storybook/react';
+import ConfirmModal from './ConfirmModal';
+import React, { useState } from 'react';
+import { action } from '@storybook/addon-actions';
+import {
+  ButtonDropdown,
+  DropdownItem,
+  DropdownMenu,
+  DropdownToggle
+} from 'reactstrap';
+
+const disclaimer = (
+  <>
+    <p className="mt-3">
+      <strong>Note:</strong> In most cases, you&apos;ll want to use the{' '}
+      <code>ConfirmButton</code> component instead. ConfirmButton offers an
+      easy-to-use button that triggers this dialog.
+    </p>
+    <p>
+      When it is not possible to trigger the action through a button, such as
+      when using a ButtonDropdown, you can use this component directly instead.
+    </p>
+  </>
+);
+
+storiesOf('core|ConfirmModal', module)
+  .addParameters({ component: ConfirmModal })
+  .add('in dropdown', () => {
+    const [isDropdownMenuOpen, setDropdownMenuOpen] = useState(false);
+    const [isConfirmModalOpen, setConfirmModalOpen] = useState(false);
+
+    return (
+      <div className="text-center mt-3">
+        <ConfirmModal
+          isOpen={isConfirmModalOpen}
+          onClose={() => setConfirmModalOpen(false)}
+          onSave={() => {
+            setConfirmModalOpen(false);
+            action('confirm clicked')();
+          }}
+          modalText="Are you sure to delete this user? This operation cannot be undone!"
+        />
+        <ButtonDropdown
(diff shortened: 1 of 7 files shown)


src/core/ConfirmModal/ConfirmModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+ * The main use case is a delete action, which we want the user to confirm before proceeding.
+ *
+ * Note: In most cases, you'll want to use {@link ConfirmButton} instead. That component offers an easy-to-use button that triggers this dialog.
+ * However, if it is not possible to trigger the action through a button, such as when using a ButtonDropdown, you can use this component directly instead.
+ */
+export default function ConfirmModal({
+  isOpen,
+  onClose,
+  onSave,
+  modalHeader,
```
MrHus, line 59: Deze `label` noemen zodat deze gelijk is aan `OpenCloseModal`

arjanvlek: The prop has been renamed to `label`.

src/core/ConfirmModal/ConfirmModal.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+import {
+  ButtonDropdown,
+  DropdownItem,
+  DropdownMenu,
+  DropdownToggle
+} from 'reactstrap';
+
+storiesOf('core|ConfirmModal', module)
+  .addParameters({ component: ConfirmModal })
+  .add('In button dropdown item - default texts', () => {
```
MrHus, line 14: Change the story name to "in dropdown"

src/core/ConfirmModal/ConfirmModal.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <DropdownMenu>
+            <DropdownItem onClick={() => setConfirmModalOpen(true)}>
+              Delete user
+            </DropdownItem>
+          </DropdownMenu>
+        </ButtonDropdown>
+      </div>
+    );
+  })
+  .add('In button dropdown item - custom text', () => {
```
MrHus, line 43: Change name to "with custom text"

src/core/ConfirmModal/ConfirmModal.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+import React, { useState } from 'react';
+import { action } from '@storybook/addon-actions';
+import {
+  ButtonDropdown,
+  DropdownItem,
+  DropdownMenu,
+  DropdownToggle
+} from 'reactstrap';
+
+storiesOf('core|ConfirmModal', module)
```
MrHus, line 12: Please add the note from the `ConfirmModal` also as text in both stories. This way the user will that the `ConfirmButton` exists as well when viewing the story.

See the Popover / TextEditor stories for an example.

src/core/ConfirmModal/ConfirmModal.stories.tsx
arjanvlek: Story name has been changed

src/core/ConfirmModal/ConfirmModal.stories.tsx
arjanvlek: Story name has been changed

src/core/ConfirmModal/ConfirmModal.stories.tsx
arjanvlek: I've added a disclaimer based on the note from the `ConfirmModal` component

MrHus (approved):

PR #521, opened 2021-01-29
improvement: option to hide clear button

### LINKED ISSUE #502: Ability to remove 'clear' buttons by using a prop in form components.
Should be possible for the following components:

1. Icon
2. Color
3. Modal picker multiple
4. Modal picker single, render as tag.

gidomanders:
Sometimes a field is required and should work like a radio button, where
it is not possible to clear the field, you have to select at least one
option. It would be nice to be able to use all form components that way,
but for that we need an option whether to display the clear button.

Added canClear option to form components to be able to prevent users
from clearing the form element.

Closes #502

### DIFF
--- a/src/form/ColorPicker/ColorPicker.test.tsx
+++ b/src/form/ColorPicker/ColorPicker.test.tsx
@@ -6,9 +6,10 @@ import { SketchPicker } from 'react-color';
 import Popover from '../../core/Popover/Popover';
 
 import ColorPicker from './ColorPicker';
+import TextButton from '../../core/TextButton/TextButton';
 
 describe('Component: ColorPicker', () => {
-  function setup({ value, hasIcon }: { value?: string; hasIcon?: boolean }) {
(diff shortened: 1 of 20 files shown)


MrHus (changes requested):
You should make the ValuePicker use it as well.

src/form/ColorPicker/ColorPicker.test.tsx
```diff
@@ -46,6 +47,12 @@ describe('Component: ColorPicker', () => {
 
       expect(toJson(colorPicker)).toMatchSnapshot();
     });
+
+    test('without clear button', () => {
+      const { colorPicker } = setup({ value: '#cdcdcd', canClear: false });
+
+      expect(toJson(colorPicker)).toMatchSnapshot();
```
MrHus, line 54: Can you instead of making snapshots everywhere simply check if the clear button is not present.

MrHus (approved):

PR #527, opened 2021-02-19
improvement: label for Toggle

### LINKED ISSUE #506: Give Toggle a label prop.
When the label is clicked the toggle should also move. 

The label should appear on the left of the toggle: https://ux.stackexchange.com/questions/113633/should-a-toggle-switchs-label-be-on-its-left-or-right

The story should use the label instead of a custom text followed by a br element.

gidomanders:
The toggle is most useful when there's a label with it, but now
developers have to add the text next to it themselves.

Added optional label to the Toggle component.

Closes #506

### DIFF
--- a/src/core/Toggle/Toggle.tsx
+++ b/src/core/Toggle/Toggle.tsx
@@ -34,6 +34,11 @@ type Props = {
    * Optional callback for when the Toggle is blurred.
    */
   onBlur?: React.FocusEventHandler;
+
+  /**
+   * Optional label to display next to the toggle.
+   */
+  label?: React.ReactNode;
 };
 
 /**
@@ -47,7 +52,8 @@ export default function Toggle({
   id,
   value = false,
   onChange,
-  onBlur
+  onBlur,
+  label
 }: Props) {
   const toggleClasses = classNames(
     'toggle-container',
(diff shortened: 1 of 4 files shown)


src/core/Toggle/Toggle.scss
```diff
@@ -1,3 +1,7 @@
+.toggle-label {
+  margin-right: 0.5rem;
```
MrHus, line 2: Try to solve this using bootstrap margin classes.

src/core/Toggle/Toggle.tsx
```diff
@@ -56,14 +62,21 @@ export default function Toggle({
...
-        onBlur={onBlur}
-      />
-    </span>
+    <>
+      {label ? (
+        <span className="toggle-label" onClick={() => onChange(!value)}>
+          {label}
+        </span>
+      ) : null}
+      <span className={toggleClasses}>
```
MrHus, line 71: The `className` is always added to the first element which is returned in 42.ui. In this case I would expect the label to be inside the span and not a sibling element.

MrHus (changes requested):

MrHus (approved):

PR #528, opened 2021-02-19
feature: modal with suspense

### LINKED ISSUE #505: Modal component with suspense
Right now we have a OpenCloseModal, we've seen how ever that it is often not enough. 
We want to create a Modal abstraction which can accept a the header / footer and body
and which will suspend by default.

Replaces Modal from react-strap.

gidomanders:
Right now we have a OpenCloseModal, we've seen how ever that it is often
not enough. We want to create a Modal abstraction which can accept a
header / footer and body and which will suspend by default.

Added new Modal component with optional header and footer and a suspense
surrounding the body.
Changed OpenCloseModal to extend the new Modal component.
Deprecated isOpen property in all modal components. The component should
not be responsible for when it is displayed or any fallback, the 
developer should only render the component when it should be rendered.

Closes #505


src/core/Modal/Modal.tsx
```diff
@@ -0,0 +1,100 @@
...
+} from 'reactstrap';
+import { BootstrapSize } from '../types';
+import { useBodyFixOnModalClose } from '../useBodyFixOnModalClose/useBodyFixOnModalClose';
+import Loading from '../Loading/Loading';
+
+type Props = {
+  /**
+   * Whether or not the modal is open.
+   */
+  isOpen: boolean;
```
MrHus, line 16: Remove this `isOpen` prop, make the Modal always opened on render. The reason to do this is so that modal and their content are not loaded unless open. This way we force the developer to do this, optimising modals along the way.

src/core/Modal/Modal.tsx
```diff
@@ -0,0 +1,100 @@
...
+  return (
+    <ReactstrapModal
+      wrapClassName={stickyFooter ? 'sticky-modal' : undefined}
+      isOpen={isOpen}
+      toggle={() => onClose()}
+      size={size}
+      className={className}
+    >
+      {header ? (
+        <ModalHeader toggle={() => onClose()}>{header}</ModalHeader>
```
MrHus, line 92: Also add `modalHeaderClassName` here.

src/core/Modal/Modal.tsx
```diff
@@ -0,0 +1,100 @@
...
+      ) : null}
+      <ModalBody className={modalBodyClassName}>
+        <Suspense fallback={<Loading />}>{children}</Suspense>
+      </ModalBody>
+      {footer ? <ModalFooter>{footer}</ModalFooter> : null}
```
MrHus, line 97: Also add `modaFooterClassName` here.

src/core/Modal/Modal.scss
```diff
@@ -0,0 +1,17 @@
+.sticky-modal {
```
MrHus, line 1: Is the OpenCloseModal.scss still needed now?

src/core/OpenCloseModal/OpenCloseModal.tsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
-import { Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
 import Button from '../Button/Button';
 import { t } from '../../utilities/translation/translation';
 import { BootstrapSize } from '../types';
 import { useBodyFixOnModalClose } from '../useBodyFixOnModalClose/useBodyFixOnModalClose';
 import { IconType } from '../Icon';
+import { Modal } from '../Modal/Modal';
 
 type Text = {
```
MrHus, line 9: We should deprecate the OpenCloseModal isOpen property and remove it in 4.0.0.

```tsx
/**
   * Whether or not the modal is open.
   * @deprecated Please do not use the isOpen boolean to control whether
   * or not the modal is opened. Instead always set isOpen to `true`
   * and only render the OpenCloseModal when it should be rendered.
   * In version 4.0.0 the isOpen property will be removed.
   */
  isOpen: boolean;
```

Also update the OpenCloseModal stories showing the desired way of doing things, plus add a message in the story book about the deprecation as well.

src/core/Modal/Modal.tsx
```diff
@@ -0,0 +1,109 @@
...
+  return (
+    <ReactstrapModal
+      wrapClassName={stickyFooter ? 'sticky-modal' : undefined}
+      isOpen={true}
+      toggle={() => onClose()}
+      size={size}
+      className={className}
+    >
+      {header ? (
+        <ModalHeader toggle={() => onClose()} className={modalHeaderClassName}>
```
MrHus, line 97: Suspend the header as well.

```tsx
{header ? (
        <ModalHeader toggle={() => onClose()} className={modalHeaderClassName}>
           <Suspense fallback={<Loading />}>{header}</Suspense>
        </ModalHeader>
      ) : null}
```

src/core/Modal/Modal.tsx
```diff
@@ -0,0 +1,109 @@
...
+          {header}
+        </ModalHeader>
+      ) : null}
+      <ModalBody className={modalBodyClassName}>
+        <Suspense fallback={<Loading />}>{children}</Suspense>
+      </ModalBody>
+      {footer ? (
```
MrHus, line 104: The footer too:

```tsx
{footer ? (
        <ModalFooter className={modalFooterClassName}>
          <Suspense fallback={<Loading />}>{footer}</Suspense>
        </ModalFooter>
      ) : null}
```

src/core/Modal/Modal.tsx
```diff
@@ -0,0 +1,109 @@
...
+  modalFooterClassName?: string;
+
+  /**
+   * The content of the modal body.
+   */
+  children: React.ReactNode;
+};
+
+/**
+ * Modal is a dialog that overlays the website. The dialog is often used to
```
MrHus, line 71: Also add that the developer should use this modal over react-straps modal because it has a suspense.

MrHus (approved):

PR #534, opened 2021-02-26
feature: info tooltip

### LINKED ISSUE #533: Info tooltip
You sometimes want to explain the use of a certain field for people that use a form for the first time, but you don't want to display that information all the time because it increases the size of the form and therefore decreases its useability. A tooltip might be the perfect solution in that case.

gidomanders:
You sometimes want to explain the use of a certain field for people that
use a form for the first time, but you don't want to display that
information all the time because it increases the size of the form and
therefore decreases its usability. A tooltip might be the perfect
solution in that case.

Added InfoTooltip compenent to display an info icon with a tooltip.

Closes #533

### DIFF
--- a/src/core/InfoTooltip/InfoTooltip.tsx
+++ b/src/core/InfoTooltip/InfoTooltip.tsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import { Icon } from '../../core/Icon';
+import Tooltip from '../../core/Tooltip/Tooltip';
+
+type Props = {
+  /**
+   * The content of the tooltip.
+   */
+  tooltip: React.ReactNode;
+
+  /**
+   * Optionally the size of the icon.
+   * Defaults to 16.
+   */
+  size?: number;
+
+  /**
+   * Optional extra CSS class you want to add to the component.
+   * Useful for styling the component.
+   */
+  className?: string;
+};
+
+/**
+ * InfoTooltip is an info icon that displays a tooltip on mouseover.
+ */
+export function InfoTooltip(props: Props) {
+  const { tooltip, size = 16, className } = props;
+  return (
+    <Tooltip content={tooltip} className={className}>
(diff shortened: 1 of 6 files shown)


src/core/InfoTooltip/InfoTooltip.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Icon } from '../../core/Icon';
+import Tooltip from '../../core/Tooltip/Tooltip';
+
+type Props = {
+  /**
+   * The content of the tooltip.
+   */
+  tooltip: string;
```
MrHus, line 9: Can this become a ReactNode?

src/core/InfoTooltip/InfoTooltip.tsx
```diff
@@ -0,0 +1,31 @@
...
+  size?: number;
+
+  /**
+   * Optional extra CSS class you want to add to the component.
+   * Useful for styling the component.
+   */
+  className?: string;
+};
+
+export function InfoTooltip(props: Props) {
```
MrHus, line 24: Add some comments on when to use this component

MrHus (changes requested):

src/form/LabelWithTooltip/LabelWithTooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Optionally the id of the form element the label is for.
+   * If this is provided, clicking the label will focus the
+   * form element.
+   */
+  htmlFor?: string;
+};
+
+export function LabelWithTooltip({ name, tooltip, size, htmlFor }: Props) {
```
MrHus, line 29: Missing docs here

MrHus (changes requested):

MrHus (approved):

MrHus (approved):

PR #538, opened 2021-02-26
feature: plain text form control

### LINKED ISSUE #537: Plain text form control
In forms you sometimes want to display a value the user is not allowed to edit. The value should have the same dimensions as an input in that case, especially when the form is spread over multiple columns.

gidomanders:
In forms you sometimes want to display a value the user is not allowed
to edit. The value should have the same dimensions as an input in that
case, especially when the form is spread over multiple columns.

Added PlainTextFormControl component.

Closes #537

### DIFF
--- a/src/form/PlainTextFormControl/PlainTextFormControl.stories.tsx
+++ b/src/form/PlainTextFormControl/PlainTextFormControl.stories.tsx
@@ -1,0 +1,12 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { PlainTextFormControl } from './PlainTextFormControl';
+import { Form } from 'reactstrap';
+
+storiesOf('Form/PlainTextFormControl', module).add('basic', () => {
+  return (
+    <Form>
+      <PlainTextFormControl label="Organisation">42 BV</PlainTextFormControl>
+    </Form>
+  );
+});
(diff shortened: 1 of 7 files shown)


src/form/PlainTextFormControl/PlainTextFormControl.tsx
```diff
@@ -0,0 +1,26 @@
+import { FormGroup, Label } from 'reactstrap';
+import React from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  label: React.ReactNode;
```
MrHus, line 6: Missing doc comments on each prop

src/form/PlainTextFormControl/PlainTextFormControl.tsx
```diff
@@ -0,0 +1,26 @@
...
+  children: React.ReactNode;
+  valueClassName?: string;
+  labelClassName?: string;
+};
+
+export function PlainTextFormControl({
```
MrHus, line 12: Explanation on when to use it.
Also mention the evils of a disabled form element and that this is the replacement component for such a scenario.

src/form/PlainTextFormControl/PlainTextFormControl.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+storiesOf('Form/PlainTextFormControl', module)
+  .add('basic', () => {
+    return (
+      <Form>
+        <PlainTextFormControl label="Organisation">42 BV</PlainTextFormControl>
+      </Form>
+    );
+  })
+  .add('in a form', () => {
```
MrHus, line 15: Remove this story, add a PlainTextFormControl to the form example. Make it the first item there and call it `Key`. Add a tooltip "The 'Key' represents a part of the entity which can only be created once on create but not on edit.

MrHus (approved):

PR #554, opened 2021-03-19
feature: favorite icon

### LINKED ISSUE #551: Favorite icon
In applications where the user can add something to a favorites list, often a star icon is used to display whether the user added the item to the favorites list or not. Clicking the icon toggles whether the item is on the favorites list or not.

gidomanders:
In applications where the user can add something to a favorites list,
often a star icon is used to display whether the user added the item to
the favorites list or not. Clicking the icon toggles whether the item is
on the favorites list or not.

Added FavoriteIcon component.

Closes #551


src/core/FavoriteIcon/FavoriteIcon.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import classNames from 'classnames';
+import { useHover } from '../useHover/useHover';
+import Icon from '../Icon/Icon';
+
+type FavoriteIconProps = {
+  /**
+   * Whether or not the item is favorited
+   */
+  isFavorite: boolean;
```
MrHus, line 10: This should be a form element. We should use value instead of isFavorite, and onChange instead of onToggle.

src/index.ts
```diff
@@ -44,6 +44,7 @@ export { SuccessIcon } from './core/SuccessIcon/SuccessIcon';
 export { OrSeparator } from './core/OrSeparator/OrSeparator';
 export { AttributeList } from './core/lists/AttributeList/AttributeList';
 export { AttributeView } from './core/lists/AttributeView/AttributeView';
+export { FavoriteIcon } from './core/FavoriteIcon/FavoriteIcon';
```
MrHus, line 47: Expose useHover as well it is a great hook to have as a developer

src/core/useHover/useHover.test.tsx
```diff
@@ -0,0 +1,24 @@
+import { useHover } from './useHover';
+import React from 'react';
+
+describe('useHover', () => {
+  describe('events', () => {
```
MrHus, line 5: The describe events can go here it does not need the extra nesting

src/core/FavoriteIcon/FavoriteIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Optional extra CSS class you want to add to the component.
+   * Useful for styling the component.
+   */
+  className?: string;
+};
+
+/**
+ * FavoriteIcon displays a heart icon to show the user
```
MrHus, line 25: It displays a star icon...

src/core/FavoriteIcon/FavoriteIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+  onToggle,
+  className = ''
+}: FavoriteIconProps) {
+  const [hover, hoverEvents] = useHover();
+
+  return (
+    <div {...hoverEvents}>
+      <Icon
+        color={isFavorite || hover ? 'primary' : 'secondary'}
+        className={classNames('favorite-icon', className)}
```
MrHus, line 40: I'm missing the:

```
.favorite-icon:hover {
  color: $primary-color;
}
```
scss.

src/core/FavoriteIcon/FavoriteIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Optional extra CSS class you want to add to the component.
+   * Useful for styling the component.
+   */
+  className?: string;
+};
+
+/**
+ * FavoriteIcon displays a heart icon to show the user
+ * whether an item is favorite or not. The component
```
MrHus, line 26: The commit message references a heart as well

src/core/FavoriteIcon/FavoriteIcon.tsx
gidomanders: The problem is that the primary-color variable is some custom variable in Tingit. It is not a bootstrap variable. But instead of using hover events to change the color property for the icon, I could use scss to change the color to $primary on hover.

gidomanders: Also, that CSS doesn't work. The icon gets a `text-secondary` or `text-primary` class based on the value, which uses `!important`. So without using that as well, the color will be overwritten by these classes anyway.
Using the hover events we change the color property of the Icon to primary, so that should work without any additional CSS.

MrHus (approved):

PR #559, opened 2021-03-19
feature: Card with OpenClose

### LINKED ISSUE #558: Card with OpenClose
Now the OpenClose component is introduced, we could use that in a collapsable card, which is often used to hide details.

gidomanders:
Now the OpenClose component is introduced, we could use that in a
collapsable card, which is often used to hide details.

Added CardOpenClose component.

Closes #558

### DIFF
--- a/src/core/CardOpenClose/CardOpenClose.tsx
+++ b/src/core/CardOpenClose/CardOpenClose.tsx
@@ -1,0 +1,61 @@
+import React from 'react';
+import { OpenClose } from '../OpenClose/OpenClose';
+import { Card, CardHeader } from 'reactstrap';
+import classNames from 'classnames';
+
+type Props = {
+  /**
+   * The important details you always want to be displayed.
+   */
+  header: React.ReactNode;
+
+  /**
+   * The content that might be hidden in the collapsable body.
+   */
+  content: () => React.ReactNode;
+
+  /**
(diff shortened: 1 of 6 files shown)


src/core/CardOpenClose/CardOpenClose.tsx
```diff
@@ -0,0 +1,48 @@
...
+};
+
+/**
+ * CardOpenClose is a collapsable bootstrap card that you can use to only display
+ * important details and hide other details in a collapsable body to prevent extra
+ * long pages that cause the user to scroll a lot.
+ */
+export function CardOpenClose({ header, children, isOpen, toggle }: Props) {
+  return (
+    <Card className="my-2">
```
MrHus, line 35: I'm missing a `className` prop all components that we expose should allow customization via a className property, which is set on the first element.

src/core/CardOpenClose/CardOpenClose.tsx
```diff
@@ -0,0 +1,61 @@
...
+type Props = {
+  /**
+   * The important details you always want to be displayed.
+   */
+  header: React.ReactNode;
+
+  /**
+   * The content might be hidden in the collapsable body.
+   */
+  children: React.ReactNode;
```
MrHus, line 15: This should be a function for performance reasons.

MrHus (approved):

PR #566, opened 2021-04-13
feat (Typeahead, ModalPicker): Configurable page size

arjanvlek:
Added the ability to specify the number of suggestions for
`TypeaheadSingle` and `TypeaheadMultiple`. Also added the ability to
show all predefined options matching the input. One of our applications
needs to show as much suggestions as possible, as users need to know
which values are supported.

Added the ability to change the page size of `ModalPickerMultiple` and
`ModalPickerSingle`, for some applications it is desired to have more
than 10 items per page on screen.

### DIFF
--- a/src/form/ModalPicker/multiple/ModalPickerMultiple.tsx
+++ b/src/form/ModalPicker/multiple/ModalPickerMultiple.tsx
@@ -50,6 +50,26 @@ export type Props<T> = Omit<
     canSearch?: boolean;
 
     /**
+     * Optionally specify the number of options to show / fetch per
+     * page in the modal.
+     *
+     * When `options` is an array, it will determine how many options
+     * will be displayed per page.
+     *
+     * When `options` is a fetcher, it will determine how many options
+     * are requested through the fetcher as the `page` parameter.
+     * This means that when you set the pageSize to `100` that
+     * `100` items are fetched from the back-end. Beware of
+     * performance issues when setting the value too high.
+     *
+     * Beware that setting the page size too high will cause the UX
+     * to deteriorate on smaller screens!
+     *
+     * Defaults to `10`.
+     */
+    pageSize?: number;
+
+    /**
      * Optionally an add button to display in the Modal. Can
      * be used to dynamically add an option which was not there
      * before.
@@ -105,6 +125,7 @@ export default function ModalPickerMultiple<T>(props: Props<T>) {
     onChange,
     addButton,
     canSearch = true,
+    pageSize = 10,
     options,
     keyForOption,
     isOptionEqual,
@@ -133,7 +154,7 @@ export default function ModalPickerMultiple<T>(props: Props<T>) {
     reloadOptions,
     pageNumber,
     query,
-    size: 10,
+    size: pageSize,
     optionsShouldAlwaysContainValue: false
   });
 
--- a/src/form/ModalPicker/single/ModalPickerSingle.stories.tsx
+++ b/src/form/ModalPicker/single/ModalPickerSingle.stories.tsx
@@ -41,6 +41,31 @@ storiesOf('Form/ModalPicker/ModalPickerSingle', module)
       </Form>
     );
   })
+  .add('show 5 options per page', () => {
+    const [value, setValue] = useState<Province | undefined>(undefined);
+
+    return (
+      <Form>
+        <ModalPickerSingle<Province>
+          id="province"
+          label="Province"
+          placeholder="Please select your province"
+          options={provinces()}
+          labelForOption={(province) => province.label}
+          value={value}
+          onChange={setValue}
+          pageSize={5}
+        />
+
+        {value ? <p>Your chosen province is: {value.label}</p> : null}
+
+        <p>
+          Note: Beware of performance issues when setting the page size too
+          high, that will cause the UX to deteriorate on smaller screens!
+        </p>
+      </Form>
+    );
+  })
   .add('async options', () => {
     const [value, setValue] = useState<Province | undefined>(provinces()[0]);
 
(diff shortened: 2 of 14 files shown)


src/form/Typeahead/single/TypeaheadSingle.tsx
```diff
@@ -67,7 +83,10 @@ export default function TypeaheadSingle<T>(props: Props<T>) {
     reloadOptions,
     query,
     pageNumber: 1,
-    size: 10,
+    size:
+      Array.isArray(options) && showAllSuggestionsInDropdown
```
MrHus, line 87: The `showAllSuggestionsInDropdown` is not necessary take a look at `RadioGroup`, `Select` and `CheckboxMultipleSelect` they simply when the options is an array use the length of the array as the size.

src/form/ModalPicker/multiple/ModalPickerMultiple.tsx
```diff
@@ -49,6 +49,12 @@ export type Props<T> = Omit<
      */
     canSearch?: boolean;
 
+    /**
+     * Optionally specify the number of elements to show per page in the modal.
```
MrHus, line 53: I Question if we really want to be able do do this. Fetching many options will be a performance hit, and showing to many options will make the modal too large which is detrimental to the user experience. 

Can you give me an example of when this is useful in your application?

src/form/Typeahead/single/TypeaheadSingle.tsx
arjanvlek: I have removed the `showAllSuggestionsInDropdown` prop with the latest patch.

src/form/ModalPicker/multiple/ModalPickerMultiple.tsx
arjanvlek: While I agree that fetching too many options degrades performance and UX, a customer of ours has requested to increase the options slightly above 10. 

As for the typeahead, we are using predefined options, so performance is not a concern.

src/form/ModalPicker/single/ModalPickerSingle.stories.tsx
```diff
@@ -41,6 +41,26 @@ storiesOf('Form/ModalPicker/ModalPickerSingle', module)
...
+          placeholder="Please select your province"
+          options={provinces()}
+          labelForOption={(province) => province.label}
+          value={value}
+          onChange={setValue}
+          pageSize={5}
+        />
+
+        {value ? <p>Your chosen province is: {value.label}</p> : null}
+      </Form>
```
gidomanders, line 61: Please also add a note about high page size deteriorating UX on smaller screens (same note as in the property comments)

src/form/Typeahead/multiple/TypeaheadMultiple.stories.tsx
```diff
@@ -69,6 +69,34 @@ storiesOf('Form/Typeahead/JarbTypeaheadMultiple', module)
...
+          pageSize={5}
+        />
+
+        {value ? (
+          <p>
+            Your chosen provinces are:{' '}
+            {value.map((province) => province.label).join(', ')}
+          </p>
+        ) : null}
+      </Form>
```
gidomanders, line 97: Here a similar note about high page size

src/form/Typeahead/multiple/TypeaheadMultiple.tsx
```diff
@@ -21,7 +21,24 @@ import { useId } from '../../../hooks/useId/useId';
...
+     * in the dropdown.
+     *
+     * When `options` is an array, all options will always be shown
+     * and this prop has no effect.
+     *
+     * When `options` is a fetcher, it will determine how many options
+     * are requested through the fetcher as the `page` parameter.
+     * This means that when you set the pageSize to `100` that
+     * `100` items are fetched from the back-end. Beware of
+     * performance issues when setting the value too high.
```
gidomanders, line 36: Please add the UX deterioration here too to make the comment exactly the same on all components. Maybe this property could even be moved to the FieldCompatibleWithPredeterminedOptions type?

src/form/Typeahead/single/TypeaheadSingle.stories.tsx
```diff
@@ -55,6 +55,28 @@ storiesOf('Form/Typeahead/JarbTypeaheadSingle', module)
...
+          placeholder="Please select your province"
+          options={provinceFetcher}
+          labelForOption={(province) => province.label}
+          value={value}
+          onChange={setValue}
+          pageSize={5}
+        />
+
+        {value ? <p>Your chosen province is: {value.label}</p> : null}
+      </Form>
```
gidomanders, line 77: Please add the aforementioned note here too

src/form/Typeahead/single/TypeaheadSingle.test.tsx
```diff
@@ -90,6 +93,34 @@ describe('Component: TypeaheadSingle', () => {
...
+
+      expect(toJson(typeaheadSingle)).toMatchSnapshot(
+        'Component: TypeaheadSingle => ui => async with a custom pageSize of 2 options in the dropdown'
+      );
+    });
+
+    test('async with the default pageSize of 10 options in the dropdown', () => {
+      const { typeaheadSingle } = setup({
+        isAsync: true,
+        pageSize: undefined
```
gidomanders, line 113: You don't have to specify undefined parameters as the default is undefined

src/form/Typeahead/single/TypeaheadSingle.tsx
```diff
@@ -17,7 +17,24 @@ import { optionToTypeaheadOption } from '../utils';
...
+     * in the dropdown.
+     *
+     * When `options` is an array, all options will always be shown
+     * and this prop has no effect.
+     *
+     * When `options` is a fetcher it will determine how many options
+     * are requested through the fetcher as the `page` parameter.
+     * This means that when you set the pageSize to `100` that
+     * `100` items are fetched from the back-end. Beware of
+     * performance issues when setting the value too high.
```
gidomanders, line 32: Same here about the deteriorating UX

gidomanders (changes requested):

MrHus (approved):

gidomanders (approved):

PR #568, opened 2021-04-16
improvement: search input clear button

### LINKED ISSUE #565: SearchInput add clear button option
It would be nice if the SearchInput showed a "X" button to clear the content once it has a value.

gidomanders:
It would be nice if the SearchInput showed a "X" button to clear the
content once it has a value.

Changed input type to search and added some styling to the browsers
search cancel button.

Closes #565

### DIFF
--- a/src/core/SearchInput/SearchInput.scss
+++ b/src/core/SearchInput/SearchInput.scss
@@ -1,0 +1,12 @@
+.search-input input[type='search']::-webkit-search-cancel-button {
+  display: none;
+}
+
+.cancel-search {
+  margin-top: -24px;
+  margin-right: 16px;
+  position: relative;
+  float: right;
+  top: 31px;
+  z-index: 4;
+}
--- a/src/core/SearchInput/SearchInput.tsx
+++ b/src/core/SearchInput/SearchInput.tsx
@@ -118,6 +118,13 @@ export type Props = ModifiedInputProps & {
    * Optional size you want to give the icon.
    */
   size?: BootstrapSize;
+
+  /**
+   * Whether or not to show a "clear" button.
+   *
+   * Defaults to `true`
+   */
+  canClear?: boolean;
 };
 
 /**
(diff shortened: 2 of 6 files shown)


src/core/SearchInput/SearchInput.scss
```diff
@@ -0,0 +1,12 @@
+input[type='search'].search-input,
+.search-input input[type='search'] {
+  &::-webkit-search-cancel-button {
+    background: $input-group-addon-color;
+    -webkit-mask: url(https://pro.fontawesome.com/releases/v5.10.0/svgs/solid/times.svg)
```
MrHus, line 5: Not a fan of using font-awesome for this. Perhaps an `<Icon icon="close" />` which is moved through some position logic to the correct place?

Also i think this might be payed only because it is the `pro` version.

src/core/SearchInput/SearchInput.tsx
```diff
@@ -198,10 +199,13 @@ export default function SearchInput(props: Props) {
     return <Input className={className} {...inputProps} />;
   }
 
-  const searchInputWrapper = children ? (
-    <>{children(getInput(), { setValue })}</>
-  ) : (
-    getInput()
+  const searchInputWrapper = (
+    <div className="search-input position-relative">
```
MrHus, line 203: Why does the search-input have a position relative?

src/core/SearchInput/SearchInput.tsx
```diff
@@ -198,10 +199,13 @@ export default function SearchInput(props: Props) {
     return <Input className={className} {...inputProps} />;
   }
 
-  const searchInputWrapper = children ? (
-    <>{children(getInput(), { setValue })}</>
-  ) : (
-    getInput()
+  const searchInputWrapper = (
```
MrHus, line 202: The `input--with-external-value` story is broken. The Icon should be rendered before rendering the children:

```tsx
<div className="search-input">
      {inputRef.current?.value ? (
        <Icon icon="close" className="cancel-search" />
      ) : null}
      {children ? <>{children(getInput(), { setValue })}</> : getInput()}
    </div>
```

Then giving it the following SCSS worked for me:

```scss
.cancel-search {
  margin-top: -25px;
  margin-right: 5px;
  pointer-events: none;
  position: relative;
  float: right;
  top: 30px;
  z-index: 3;
}
```

MrHus (approved):

PR #570, opened 2021-04-16
feature: tabs

### LINKED ISSUE #569: Tabs
It would be nice to have tabs where the content and navigation item are together so you can disable a tab or specify if a user is allowed to see a tab in one place without having to put the code all over the place. The content should also be suspended to improve performance.

gidomanders:
It would be nice to have tabs where the content and navigation item are
together so you can disable a tab or specify if a user is allowed to see
a tab in one place without having to put the code all over the place.
The content should also be suspended to improve performance.

Added Tabs component.

Closes #569

### DIFF
--- a/src/core/Tabs/Tab/Tab.tsx
+++ b/src/core/Tabs/Tab/Tab.tsx
@@ -1,0 +1,91 @@
+import React from 'react';
+import { NavItem, NavLink } from 'reactstrap';
+import { Icon, IconType } from '../../Icon';
+import { Color } from '../../types';
+import { alwaysTrue } from '../../../form/utils';
+import classNames from 'classnames';
+
+export type Props = {
+  /**
+   * The label to display in the tab navigation.
+   */
+  label: string;
+
+  /**
+   * Callback that should activate the tab.
+   *
+   * We highly recommend adding the tab as a URL parameter,
+   * especially if the inactive tabs are not hidden by CSS.
+   */
+  onClick: () => void;
+
+  /**
+   * Callback to render the tab content.
+   */
+  children: () => React.ReactNode;
+
+  /**
+   * Whether or not the tab content should be displayed.
+   */
+  active: boolean;
+
+  /**
+   * Optional boolean to determine if the tab navigation item should be disabled.
+   *
+   * Defaults to false.
+   */
+  disabled?: boolean;
+
+  /**
+   * Optional callback to determine if the tab should be displayed.
+   * This is useful if you have to do permission checks to determine
+   * if the user is allowed to see the tab content.
+   *
+   * By default, tabs will always be displayed.
+   */
+  show?: () => boolean;
+
+  /**
+   * Optional icon to be displayed in front of the label in the tab navigation.
+   */
+  icon?: IconType;
+
+  /**
+   * Optional color the icon should have.
+   * This is useful when using tabs in a form and an inactive tab
+   * contains validation errors.
+   */
+  iconColor?: Color;
+};
+
+/**
+ * The Tab component displays a tab navigation item and holds the content
+ * that will only be displayed when the tab is active.
+ * This component should only be used inside the Tabs component.
+ */
+export function Tab({
+  icon,
+  label,
+  onClick,
+  active,
+  show = alwaysTrue,
+  disabled,
+  iconColor
+}: Props) {
+  if (!show()) {
+    return null;
+  }
+
+  return (
+    <NavItem className={classNames({ disabled })}>
+      <NavLink active={active} onClick={onClick} disabled={disabled}>
+        <div className="d-flex p-2 justify-content-center">
+          {icon ? (
+            <Icon icon={icon} color={iconColor} className="mr-1" />
+          ) : null}
+          {label}
+        </div>
+      </NavLink>
+    </NavItem>
+  );
+}
--- a/src/core/Tabs/TabContent/TabContent.tsx
+++ b/src/core/Tabs/TabContent/TabContent.tsx
@@ -1,0 +1,36 @@
+import { alwaysTrue } from '../../../form/utils';
+import classNames from 'classnames';
+import React from 'react';
+import { HideInactiveTabsBy } from '../Tabs';
+
+export type Props = {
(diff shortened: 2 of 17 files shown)


src/core/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+import { Tabs } from './Tabs';
+
+import { Form } from 'react-final-form';
+import { Tab, Props as TabProps } from './Tab/Tab';
+import Button from '../Button/Button';
+import { JarbInput } from '../../form/Input/Input';
+import { JarbTextarea } from '../../form/Textarea/Textarea';
+
+const tabs: TabProps[] = [
```
MrHus, line 12: Inline this into each story, this makes the stories difficult to copy paste, also this stimulates our users using arrays which is not needed.

src/core/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+    active: false,
+    // eslint-disable-next-line react/display-name
+    content: () => <p>This is suspended content</p>,
+    disabled: true,
+    show: () => true,
+    icon: 'warning',
+    iconColor: 'primary'
+  },
+  {
+    label: 'testing',
```
MrHus, line 39: Give this tab a different name

src/core/Tabs/Tab/Tab.tsx
```diff
@@ -0,0 +1,90 @@
...
+  show = alwaysTrue,
+  disabled,
+  iconColor
+}: Props) {
+  if (!show()) {
+    return null;
+  }
+
+  return (
+    <NavItem>
```
MrHus, line 79: For some reason there is no cursor pointer when hovering over the tabs.

src/core/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+storiesOf('core/Tabs', module)
+  .addParameters({ component: Tabs })
+  .addParameters({ component: Tab })
+  .add('default', () => {
+    const [activeTab, setActiveTab] = useState(0);
+    return (
+      <Tabs>
+        <Tab
+          onClick={() => setActiveTab(0)}
+          active={activeTab === 0}
```
MrHus, line 61: Make this a predefined type which is a string. This way people can easier see how to use it in combination with a url param. 

`type Tabs = 'info' | 'testing' | 'button'`.

src/core/Tabs/Tab/Tab.tsx
```diff
@@ -0,0 +1,90 @@
...
+   *
+   * We highly recommend adding the tab as a URL parameter,
+   * especially if the inactive tabs are not hidden by CSS.
+   */
+  onClick: () => void;
+
+  /**
+   * Callback to render the tab content.
+   */
+  content: () => React.ReactNode;
```
MrHus, line 24: Make this the `children` instead, make sure you use it like children in the stories as well.

```tsx
<Tab>
  {() => <p>Content here></p>}
</Tab>
```

src/core/Tabs/Tabs.scss
```diff
@@ -0,0 +1,13 @@
+.tabs {
+  .nav-item.disabled {
+    cursor: not-allowed;
+  }
+
+  .nav-link {
+    cursor: pointer;
+
+    &.active {
+      cursor: initial;
```
MrHus, line 10: Just use cursor pointer for the active one as well, even though nothing will happen.

src/core/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,221 @@
...
+import { JarbTextarea } from '../../form/Textarea/Textarea';
+
+storiesOf('core/Tabs', module)
+  .addParameters({ component: Tabs })
+  .addParameters({ component: Tab })
+  .add('default', () => {
+    type AvailableTabs = 'info' | 'testing' | 'details';
+    const [activeTab, setActiveTab] = useState<AvailableTabs>('info');
+    return (
+      <Tabs>
```
MrHus, line 19: Please wrap the stories in a `Card` which will make them a little better looking.

src/core/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  return (
+    <div className={classNames('tabs', className)}>
+      <Nav tabs fill style={{ overflowY: 'hidden' }} className="flex-nowrap">
+        {children}
+      </Nav>
+      <div
+        className={classNames('mt-3', {
+          'tab-content': hideInactiveTabsBy === 'CSS'
```
MrHus, line 75: I wonder why we do this? I think adding `tab-content` CSS class will just always work, regardless of the `hideInactiveTabsBy` state.

src/core/Tabs/TabContent/TabContent.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  if (hideInactiveTabsBy === 'excluding-from-dom' && !active) {
+    return null;
+  }
+
+  return (
+    <div
+      className={classNames({
+        'tab-pane': hideInactiveTabsBy === 'CSS',
```
MrHus, line 30: I wonder why we do this? I think adding `tab-pane` CSS class will just always work, regardless of the `hideInactiveTabsBy` state.

src/core/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,221 @@
...
+                )}
+              </Tab>
+            </Tabs>
+          </>
+        )}
+      />
+    );
+  });
+
+function sleep(ms: number) {
```
MrHus, line 217: We should put this in `story-utils.tsx`

src/core/Tabs/Tabs.tsx
gidomanders: With tab-content, the tab-pane inside will be hidden unless it has the active class. That doesn't really matter with exclude-from-dom, as long as we always use the active class, so I'll change that.

MrHus (approved):

PR #577, opened 2021-05-21
feature: bootstrap size

### LINKED ISSUE #543: Bootstrap window size
Sometimes a developer wants to display something only on desktop or only on mobile screens. For that, we need to know what size the screen is.

gidomanders:
Sometimes a developer wants to display something only on desktop or only
on mobile screens. For that, we need to know what size the screen is.

Added useBootstrapSize hook to determine what size the screen is and if
it is a mobile, tablet or desktop.

Closes #543


src/index.ts
```diff
@@ -52,6 +52,7 @@ export { Card } from './core/Card/Card';
 export { CardOpenClose } from './core/CardOpenClose/CardOpenClose';
 export { Tabs } from './core/Tabs/Tabs';
 export { Tab } from './core/Tabs/Tab/Tab';
+export { useBootstrapSize } from './core/useBootstrapSize/useBootstrapSize';
```
MrHus, line 55: I would expect this in hooks.

src/core/useBootstrapSize/useBootstrapSize.ts
```diff
@@ -0,0 +1,91 @@
...
+ *    )
+ *  }
+ *
+ *  return (
+ *    // Render for desktop
+ *  )
+ * }
+ * ```
+ */
+export function useBootstrapSize(): BootstrapInfo {
```
MrHus, line 44: I would like to see a story which demonstrates `useBootstrapSize`.

Perhaps some think which changes color based on the bootstrap size.

MrHus (approved):

PR #581, opened 2021-05-28
feature: Async action button

### LINKED ISSUE #580: Async action button
I want to have a button that triggers a request that should only be triggered once at a time, so the button should be disabled while the request is still pending and it should display a loading state.

gidomanders:
I want to have a button that triggers a request that should only be
triggered once at a time, so the button should be disabled while the
request is still pending and it should display a loading state.

Added the AsyncActionButton component.

Closes #580

### DIFF
--- a/src/core/AsyncActionButton/AsyncActionButton.tsx
+++ b/src/core/AsyncActionButton/AsyncActionButton.tsx
@@ -1,0 +1,38 @@
+import React, { useState } from 'react';
+
+import Button, { Props as ButtonProps } from '../Button/Button';
+
+export type Props = Omit<ButtonProps, 'onClick' | 'inProgress'> & {
+  /**
+   * Callback that is triggered when the button is clicked.
+   * This callback should return a request promise.
+   */
+  action: () => Promise<unknown>;
+};
+
+/**
+ * The AsyncActionButton is a button that triggers a request when it is
+ * clicked and is disabled until the request is finished. This button should
+ * be used to trigger resource-heavy actions that should not be triggered
+ * multiple times asynchronously.
+ */
+export function AsyncActionButton({ action, children, ...buttonProps }: Props) {
+  const [inProgress, setInProgress] = useState(false);
+
+  async function onClick() {
+    setInProgress(true);
+
+    try {
+      await action();
+      setInProgress(false);
+    } catch {
+      setInProgress(false);
+    }
+  }
+
+  return (
+    <Button onClick={onClick} inProgress={inProgress} {...buttonProps}>
+      {children}
+    </Button>
+  );
+}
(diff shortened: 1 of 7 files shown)


src/core/AsyncActionButton/AsyncActionButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  onFailure: (reason: any) => void;
+};
+
+/**
+ * The AsyncActionButton is a button that triggers a request when it is
+ * clicked and is disabled until the request is finished. This button should
+ * be used to trigger resource-heavy actions that should not be triggered
+ * multiple times asynchronously.
+ */
+export default function AsyncActionButton({
```
MrHus, line 29: The `AsyncActionButton` is not exported in the index barrel.

src/core/AsyncActionButton/AsyncActionButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Callback that is triggered when the button is clicked.
+   * This callback should return a request promise.
+   */
+  action: () => Promise<any>;
+
+  /**
+   * Callback that is triggered when the request succeeds.
+   */
+  onSuccess: (response: any) => void;
```
MrHus, line 15: Perhaps we were wrong about these callbacks being needed. I think it is better that the developers themselves do something in their "action" we can still tell them it is better to show a flash message.

src/core/AsyncActionButton/AsyncActionButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  onSuccess,
+  onFailure,
+  children,
+  ...buttonProps
+}: Props) {
+  const [inProgress, setInProgress] = useState(false);
+
+  function onClick() {
+    setInProgress(true);
+    action()
```
MrHus, line 40: Write this using async:

```js
async function onClick() {
  setInProgress(true);

  try {
    await action();
  } finally {
    setInProgress(false);
  }
}
```

src/core/AsyncActionButton/AsyncActionButton.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from 'react';
+
+import Button, { Props as ButtonProps } from '../Button/Button';
+
+export type Props = Omit<ButtonProps, 'onClick' | 'inProgress'> & {
+  /**
+   * Callback that is triggered when the button is clicked.
+   * This callback should return a request promise.
+   */
+  action: () => Promise<any>;
```
MrHus, line 10: If we had kept the onSuccess and the onFailure you should have still used generics for the succes and error.

Still use `action: () => Promise<unkown>;` here.

MrHus (approved):

MrHus (approved):

PR #586, opened 2021-06-21
Improvement/submit button scroll to closest error

MrHus:


src/core/SubmitButton/SubmitButton.tsx
```diff
@@ -31,28 +33,57 @@ export type Props = {
...
+  /**
+   * Optionally the Icon you want to use.
+   *
+   * Defaults to 'save'.
+   */
+  icon?: IconType;
 };
 
 /**
- * A SubmitButton is a Button with a 'save' icon and and
```
gidomanders, line 37: Double "and"

src/core/SubmitButton/useScrollToClosestError.test.ts
```diff
@@ -0,0 +1,178 @@
...
+    // Now it should trigger the debounced scroll
+    act(() => {
+      jest.advanceTimersByTime(199);
+    });
+    expect(document.querySelector).toBeCalledTimes(2);
+    expect(scrollIntoViewSpy).toBeCalledTimes(2);
+  });
+
+  test('when enabled not scroll when no element is found', () => {
+    const scrollIntoViewSpy = jest.fn();
```
gidomanders, line 83: This sentence does not flow nicely. Also maybe use it instead of test? That way you might be forced to write better explanations.

gidomanders (changes requested):

gidomanders (approved):

PR #602, opened 2021-10-08
improvement: replace React Async with React Query

### LINKED ISSUE #601: Replace react-async dependency
There is a dependency to react-async, which is used in the AsyncContent, AsyncList and AsyncPage components. That library has not been touched in a year and is prone to security issues. There are a few other libraries like react-query or swr which provide the same functionality and more.

gidomanders:
There is a dependency to React Async, which is used in the AsyncContent,
AsyncList and AsyncPage components. That library has not been touched in
a year and is prone to security issues. There are a few other libraries
like React Query or SWR which provide the same functionality and more.

Added support for React Query in Async components.

Closes #601

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "react-async": "10.0.1",
     "react-docgen-typescript-plugin": "1.0.0",
     "react-dom": "17.0.2",
+    "react-query": "3.25.1",
     "react-router-dom": "5.3.0",
     "react-test-renderer": "17.0.2",
     "reactstrap": "8.10.0",
--- a/src/core/AsyncContent/AsyncContent.stories.tsx
+++ b/src/core/AsyncContent/AsyncContent.stories.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 
 import AsyncContent from './AsyncContent';
-import { useAsync } from 'react-async';
+import { useQuery } from 'react-query';
 import { action } from '@storybook/addon-actions';
 import { ContentState, Button } from '../..';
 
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -114,9 +114,9 @@
     "np": "7.5.0",
     "prettier": "2.4.1",
     "react": "17.0.2",
-    "react-async": "10.0.1",
     "react-docgen-typescript-plugin": "1.0.0",
     "react-dom": "17.0.2",
+    "react-query": "^3.25.1",
```
SanderBenschop, line 119: Can you remove the carot please? We usually stick to explicit versions :)

src/core/AsyncContent/AsyncContent.stories.tsx
```diff
@@ -29,7 +29,7 @@ function loadingData() {
 storiesOf('core/async/AsyncContent', module)
   .addParameters({ component: AsyncContent })
   .add('when loaded', () => {
-    const state = useAsync(loadData);
+    const state = useQuery('data', loadData);
```
robert-bor, line 32: This key should be unique, see [documentation](https://react-query.tanstack.com/reference/useQuery):

Screenshot 2021-10-17 at 14 20 25

package.json
```diff
@@ -114,9 +114,9 @@
     "np": "7.5.0",
     "prettier": "2.4.1",
     "react": "17.0.2",
-    "react-async": "10.0.1",
     "react-docgen-typescript-plugin": "1.0.0",
     "react-dom": "17.0.2",
+    "react-query": "3.25.1",
```
robert-bor, line 119: By default react query will [retry all failed queries three times](https://react-query.tanstack.com/guides/query-retries). It might be good to set turn this off by default to keep the behavior as much as-is as possible.

SanderBenschop (approved):

PR #610, opened 2021-12-28
improvement: Pagination static number of items

### LINKED ISSUE #604: Pagination static width
Currently, the pagination component could change position when it is centered on a page depending on what page you are on. This should be changed to always display a static amount of buttons, so the position of the pagination component does not change when navigating between pages.

gidomanders:
Currently, the pagination component could change position when it is
centered on a page depending on what page you are on. This should be
changed to always display a static amount of buttons, so the position of
the pagination component does not change when navigating between pages.

Changed pagination to always display the same amount of buttons.

Closes #604

### DIFF
--- a/src/core/Pagination/Pagination.scss
+++ b/src/core/Pagination/Pagination.scss
@@ -2,6 +2,11 @@
   margin-bottom: 0;
 
   .page-item {
+    .page-link {
+      min-width: 38px;
+      padding: 0.5rem 0.5rem;
+    }
+
     button.disabled.page-link {
       border-radius: 0;
       background-color: $white;
--- a/src/core/Pagination/Pagination.test.tsx
+++ b/src/core/Pagination/Pagination.test.tsx
@@ -170,32 +170,37 @@ describe('pages', () => {
   });
 
   test('almost middle', () => {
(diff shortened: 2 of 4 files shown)


src/core/Pagination/Pagination.scss
```diff
@@ -2,6 +2,10 @@
   margin-bottom: 0;
 
   .page-item {
+    .page-link {
+      min-width: 46px;
```
TimQuist, line 6: Tested locally with storybook, and it (slightly) overflows horizontally on IPhone6/7/8 viewport with the prev/next arrows enabled. Without the arrows enabled the it looks fine.

Should we do something about this or should the prev/next just be disabled by the consumers on mobile viewports?

gidomanders: @TimQuist I've changed the padding and minimum width to fit better on mobile screens. It still doesn't quite fit the iPhone 5 screen, it overflows a little but the prev and next buttons are still mostly visible, and on about all other phone screens it fits great. I'd say 98% coverage is good enough.

TimQuist: Yeah that makes sense, re-tested and looks good!

TimQuist (approved):

PR #611, opened 2021-12-29
improvement: icon hover color

### LINKED ISSUE #582: hover color for icons
The BooleanIcon now accepts a color change on hover, but I would like to see that functionality on Icon and all of its wrappers as well.

gidomanders:
The BooleanIcon now accepts a color change on hover, but I would like to
see that functionality on Icon and all of its wrappers as well.

Added hoverColor to icon and all wrapping components.

Closes #582

### DIFF
--- a/src/core/Icon/Icon.stories.tsx
+++ b/src/core/Icon/Icon.stories.tsx
@@ -30,7 +30,16 @@ storiesOf('core/Icons', module)
         <Icon icon="drafts" />
         <Icon icon="drafts" disabled={true} />
         <Icon icon="timer_3" onClick={action('timer clicked')} />
-        <Icon icon="home" disabled={true} onClick={action('timer clicked')} />
+        <Icon icon="home" disabled={true} onClick={action('home clicked')} />
+      </div>
+    );
+  })
+  .add('hover color', () => {
+    return (
+      <div className="d-flex flex-column align-items-center">
+        <Icon icon="child_care" color="secondary" hoverColor="danger" onClick={action('child care clicked')} />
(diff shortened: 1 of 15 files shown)


src/core/Icon/Icon.stories.tsx
```diff
@@ -32,7 +32,16 @@ storiesOf('core/Icons', module)
...
+        <Icon icon="home" disabled={true} onClick={action('home clicked')} />
+      </div>
+    );
+  })
+  .add('hover color', () => {
+    return (
+      <div className="d-flex flex-column align-items-center">
+        <Icon icon="child_care" color="secondary" hoverColor="danger" onClick={action('child care clicked')} />
+        <Icon icon="train" color="primary" hoverColor="success" onClick={action('train clicked')} />
+        <Icon icon="home" disabled={true} color="primary" hoverColor="success" onClick={action('home clicked')} />
```
TimQuist, line 44: Is there an underlying reason to write `disable={true}`  instead of just `disabled`?

Or is it done just to follow the same style as is done everywhere else?

TimQuist (commented):
Tested locally and all Icons I could find that were not deeply nested within other components, have a `hoverColor` variant.

There's also a `Button` as `Icon` but looking at that, it might only be confusing to add a hoverColor property to that.

src/core/Icon/Icon.stories.tsx
gidomanders: It is pretty common to pass the actual boolean to the react component property. This is also a rule in our code style.

TimQuist (approved):

PR #612, opened 2021-12-29
feature: CrudTable

### LINKED ISSUE #605: CrudTable
We are using the EpicTable a lot for our entity listings and it would be a lot easier if there is a component that simplifies the EpicTable for that purpose.

gidomanders:
We are using the EpicTable a lot for our entity listings and it would be
a lot easier if there is a component that simplifies the EpicTable for
that purpose.

Added CrudTable component which wraps the EpicTable to make it easier to
display a table with create/edit/detail view and pagination.
Added CrudHeader component to simplify the display of a header with
sort and search functionality.

Closes #605

### DIFF
--- a/src/table/CrudTable/CrudTable.stories.tsx
+++ b/src/table/CrudTable/CrudTable.stories.tsx
@@ -1,0 +1,780 @@
+import React, { useEffect, useState } from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import { Card } from 'reactstrap';
+import { capitalize, every, lowerCase, startsWith } from 'lodash';
+import { Field, Form } from 'react-final-form';
+import moment from 'moment';
+import Button from '../../core/Button/Button';
+import Tag from '../../core/Tag/Tag';
+import Select from '../../form/Select/Select';
+import Input from '../../form/Input/Input';
+import ContentState from '../../core/ContentState/ContentState';
+import ConfirmButton from '../../core/ConfirmButton/ConfirmButton';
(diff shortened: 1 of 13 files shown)


src/table/CrudTable/CrudTable.stories.tsx
```diff
@@ -0,0 +1,693 @@
...
+            }}
+          </EpicDetailRow>
+          <EpicDetailRow left={columns.firstName} active={creating}>
+            {() => (
+              <EpicDetail onClose={() => setCreating(false)}>
+                <PersonForm />
+              </EpicDetail>
+            )}
+          </EpicDetailRow>
+        </CrudTable>
```
TimQuist, line 644: The example is very detailed which is good for viewing the component in the storybook, but I find the code example that you get pretty overwhelming.  Maybe it's an idea to also create a bare-bones version so people could view that and understand the 'fundamentals' of this CrudTable?

TimQuist (commented):
Tested locally (in storybook). Seems to work just fine besides the minor 'bug' in the example

src/table/CrudTable/CrudTable.stories.tsx
gidomanders: @TimQuist I've added a "simple example" story. It's a very much thinned-out version of the full example where you don't have to scroll through that much code.

TimQuist (approved):

src/table/CrudTable/CrudTable.stories.tsx
TimQuist: Tested locally and looks good

PR #615, opened 2021-12-30
chores: updated dependencies

gidomanders:


TimQuist (approved):
Tested locally (with node.js 16 LTS version) and components in storybook seem to work fine. Is there the need to test with other node.js versions?

gidomanders:
> Tested locally (with node.js 16 LTS version) and components in storybook seem to work fine. Is there the need to test with other node.js versions?

No there is no need for testing with other node versions, we have GitHub Actions to check if the build succeeds with those versions. As long as the pipeline succeeds and node v16 works well locally, it's a go.

PR #622, opened 2022-01-16
fix: icon color

### LINKED ISSUE #621: Icons in Button and Confirmbutton don't inherit color properly, possibly due to undefined hoverColor property (>3.19)
Hi there. After upgrading from `3.18` to `3.19`, we noticed our `Confirmbutton` being rendered differently than expected. `Button` is likely affected as well.

SomeFile.tsx (simplified):
```tsx
<div className="actions">
  <Button
    className="mr-1"
    color="primary"
    onClick={() => setLabel(row)}
    icon="edit"
  />
  <ConfirmButton
    className="d-inline-block"
    onConfirm={onDelete}
    dialogText="someDialog"
    icon="delete"
  />
</div>
```

Expected output (on 3.18):
```html
<div class="actions">
   <span class="button mr-1 primary d-inline-block">
      <i style="font-size: 24px;" class="icon material-icons text-primary clickable" googl="true">edit</i>
   </span>
   <div class="d-inline-block" style="display: inline;">
      <span class="button danger d-inline-block">
         <i style="font-size: 24px;" class="icon material-icons text-danger clickable">delete</i>
      </span>
   </div>
</div>
```
renders as: 
afbeelding

However, using the same exact components and properties, the html ends up like this on 3.19:
```html
<div class="actions">
   <span class="button mr-1 primary d-inline-block">
      <i style="font-size: 24px;" class="icon material-icons clickable" googl="true">edit</i>
   </span>
   <div class="d-inline-block" style="display: inline;">
      <span class="button danger d-inline-block">
         <i style="font-size: 24px;" class="icon material-icons clickable">delete</i>
      </span>
   </div>
</div>
```
which renders as black buttons (or icons rather), since the `text-danger` or `text-primary` classes are missing from the `<i>`s:
afbeelding

I think it might have something to do with [these new lines](https://github.com/42BV/ui/commit/dceb3200c62fb95c8fb61da1b5885669d8ef935e#diff-293d465e090e256b06fff1824e1587c3b55d881e45c96efb64f7df06a3223190R73-R79) in `Icon.tsx` from `@42.nl/ui`:

```tsx
const colorCssClasses = {};
if (color) {
  colorCssClasses[`text-${color}`] = (hoverColor && !hover) || disabled || onClick === undefined;
}
if (hoverColor) {
  colorCssClasses[`text-${hoverColor}`] = hover && !disabled && onClick !== undefined;
}
```

My theory is that since the `Button` components have no `hoverColor` property (which is optional in `Icon`), the `(hoverColor && !hover)` part can never be true in that case, thus the `text-*` class is never added to the icon. 

Hope this helps a little, cheers!

\- KilledMufasa (Joris van der Burgh)

gidomanders:
After the last release with the icon hover color, the non-hover color
turns out not to work correctly anymore.

Changed check to add correct color to the icon.
Added test and changed some tests to cover all possible layouts.

Closes #621

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -159,7 +159,6 @@
       "version": "7.16.7",
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.16.7.tgz",
       "integrity": "sha512-aeLaqcqThRNZYmbMqtulsetOQZ/5gbR/dWruUCJcpas4Qoyy+QeagfDsPdMrqwsPRDNxJvBlRiZxxX7THO7qtA==",
-      "dev": true,
       "dependencies": {
         "@babel/code-frame": "^7.16.7",
         "@babel/generator": "^7.16.7",
@@ -407,7 +406,6 @@
       "version": "7.16.7",
       "resolved": "https://registry.npmjs.org/@babel/helper-module-transforms/-/helper-module-transforms-7.16.7.tgz",
       "integrity": "sha512-gaqtLDxJEFCeQbYp9aLAefjhkKdjKcdh6DB7jniIGU3Pz52WAmP268zK0VgPz9hUNkMSYeH976K2/Y6yPadpng==",
-      "dev": true,
       "dependencies": {
         "@babel/helper-environment-visitor": "^7.16.7",
         "@babel/helper-module-imports": "^7.16.7",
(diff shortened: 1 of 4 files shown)


TimQuist (approved):
Saw the same when upgrading in my project, this seems to solve it.

KilledMufasa:
Nice work, thx for the quick fix!

PR #638, opened 2022-07-01
Improvement/seed 75 improve accessibility

BarryS42:
Toegankelijkheid verbeterd voor visuele beperkingen.


src/core/SearchInput/SearchInput.tsx
```diff
@@ -197,12 +197,12 @@ export default function SearchInput(props: Props) {
       return (
         <InputGroup className={className} size={size}>
           <AddonIcon icon="search" />
-          <Input {...inputProps} />
+          <Input {...inputProps} aria-label="search" />
```
gidomanders, line 200: The violation you are trying to solve here has something to do with no related label element. We have an optional label property that adds a label element (line 223). If the developer uses that property, you'd have conflicting labels. This aria-label is not translated. We might want to consider an "invisible label" feature where all form elements require the label, but the aria-label is used when the label should not be visible.

src/core/Toggle/Toggle.tsx
```diff
@@ -74,6 +74,7 @@ export default function Toggle({
         onChange={(event) => onChange(event.target.checked)}
         checked={value}
         onBlur={onBlur}
+        aria-label="Toggle"
```
gidomanders, line 77: For this one it's the same thing, we have a label property.

src/form/DateTimeInput/DateTimeInput.tsx
```diff
@@ -179,7 +179,7 @@ export default function DateTimeInput(props: Props) {
           {label}{' '}
           <span
             className={`date-time-input-format ${
-              value && hasFormatError ? 'text-danger' : 'text-muted'
+              value && hasFormatError ? 'text-danger' : ''
             }`}
```
gidomanders, line 183: It is better readable if we use `classNames` here.

```suggestion
            className={classNames('date-time-input-format', {
              'text-danger': value && hasFormatError
            })}
```

src/form/NewPasswordInput/PasswordStrength/PasswordStrength.tsx
```diff
@@ -51,6 +51,7 @@ export default function PasswordStrength(props: Props) {
           }
           value={meterWidth}
           className="mb-2"
+          title = "Progress bar"
```
gidomanders, line 54: Remove the spaces around the `=` and make the title translatable

src/table/CrudTable/components/CrudHeader/CrudHeader.tsx
```diff
@@ -120,7 +120,7 @@ export function CrudHeader<T = string>({
               onChange={(value) => {
                 // @ts-expect-error Default value type is a string
                 onSearch(value);
-              }} />
+              }}/>
```
gidomanders, line 123: Closing the tag should be happening on a new line

src/table/EpicTable/rows/EpicRow/EpicRow.tsx
```diff
@@ -4,6 +4,8 @@ export type Props = {
   /**
    * The cells of the EpicRow
    */
+  tabIndex?: number;
```
gidomanders, line 7: The comments above are related to the `children` property, so you should move this new property above the comments. Each property requires comments, so add some for the new property as well.

src/table/EpicTable/rows/EpicRow/EpicRow.tsx
```diff
@@ -29,6 +31,6 @@ export type Props = {
  * finds another EpicRow which is a header. This creates so called
  * sections.
  */
-export function EpicRow({ children }: Props) {
-  return <>{children}</>;
+export function EpicRow({ tabIndex, ...children }: Props) {
```
gidomanders, line 34: `children` should not be spread. The spread operator in this case means `children` becomes an object with all other properties.

gidomanders (changes requested):

gidomanders:
Another PR was made with a different branch name for this.

PR #639, opened 2022-07-05
Added a CopyToClipboard element

BarryS42:
Added a CopyToClipboard element, which copies the content of its child elements without HTML code.

### DIFF
--- a/docs/translations.md
+++ b/docs/translations.md
@@ -204,6 +204,9 @@ Here is a JSON notation containing all keys and default translations:
     "ERROR": {
       "TITLE": "Oops something went wrong!"
     }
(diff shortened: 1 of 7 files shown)


gidomanders:
I want to see something looking more like this (ignore the green line):
image
so with a background to make sure the user knows which part is copied. I also want a property "mode" that is either "overflow" or "wrapped". Overflow will make the text a one-liner that is horizontally scrollable left of the copy button. Wrapped will wrap the text and prevent it from disappearing behind the button.

I also want to see the text "copied!" left of the icon for a few seconds when the text is copied to notify the user the text is actually copied.

PR #640, opened 2022-07-08
feature: pagination page size dropdown

BarryS42:
It would be nice if the pagination component also has an option to 
change the page size so users can control how many items they see in a 
list.

Added page size dropdown to pagination component.

### DIFF
--- a/src/core/Pagination/Pagination.tsx
+++ b/src/core/Pagination/Pagination.tsx
@@ -9,6 +9,7 @@ import {
 
 import { Icon } from '../Icon';
 import { range } from 'lodash';
+import { Select } from '../../form/Select/Select';
 
 type Props<T> = {
   /**
@@ -20,6 +21,18 @@ type Props<T> = {
    * Called when navigation to a certain page number.
    */
   onChange: (pageNumber: number) => void;
+
+  /**
+   * Optional callback to change the page's size using a dropdown.
+   * Only if this property is provided, the dropdown will be displayed.
+   */
+  onPageSizeChange?: (size: number) => void;
+
+  /**
+   * Optional list of page sizes used for the page size dropdown.
+   * Defaults to `[5, 10, 20, 50, 100]`.
+   */
+  allowedPageSizes?: number[];
 
   /**
    * Whether or not to show the previous and next buttons.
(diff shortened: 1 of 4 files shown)


src/core/Pagination/Pagination.tsx
```diff
@@ -21,6 +22,12 @@ type Props<T> = {
    */
   onChange: (pageNumber: number) => void;
 
+  /**
+   * Called when the user changes the page's size using the dropdown.
+   * The dropdown is only visible if this is not null.
```
gidomanders, line 27:
```suggestion
   * Optional callback to change the page's size using a dropdown.
```

src/core/Pagination/Pagination.tsx
```diff
@@ -21,6 +22,12 @@ type Props<T> = {
...
+   */
+  onSizeChange?: (size: number) => void;
+
```
gidomanders, line 30: I would add an optional property to allow the developer to determine the step size of the dropdown and limit the numbers in the size dropdown. This allows for column layouts not to break when the chosen size is not a multiple of the number of columns. This also allows for some limitations to prevent performance issues. The default sizes you're using are fine, but not flexible enough.

src/core/Pagination/Pagination.tsx
```diff
@@ -47,49 +54,65 @@ type Props<T> = {
 export default function Pagination<T>({
   page,
   onChange,
+  onSizeChange,
   className,
   showPreviousAndNextButtons = true
 }: Props<T>) {
   const { first, last, totalPages } = page;
+  const current = page.number;
```
gidomanders, line 62:
```suggestion
  const { first, last, totalPages, current } = page;
```

src/core/Pagination/Pagination.tsx
```diff
@@ -47,49 +54,65 @@ type Props<T> = {
...
+        {onSizeChange ? <Select<number>
+            onChange={(size) => {
+              onSizeChange(size);
+              sizeChange(size);
+            }}
+            options={sizes}
+            labelForOption={(size) => size.toString()}
+            value={size}
+            className="col col-lg-auto select-size"
+            aria-label="Select the page's size"/> : null}
```
gidomanders, line 113: Make this property translatable

gidomanders (changes requested):

PR #641, opened 2022-07-08
improvement: accessibility

BarryS42:
We would like our components to be accessible, so they could be used in
websites that have to comply with the WCAG standards.

Changed some (text) colors to improve contrast.
Changed label property of form elements to be required.
Added property to form elements to use invisible label (aria-label).
Added a11y addon to Storybook to be able to test and display WCAG 2
compatibility.


src/core/Toggle/__snapshots__/Toggle.test.tsx.snap
```diff
@@ -49,6 +49,7 @@ exports[`Component: Toggle ui without label: Component: Toggle => ui => without
   className="toggle-container toggle-primary"
 >
   <input
+    aria-label="Toggle"
```
gidomanders, line 52: Toggle doesn't really say much, but I'll allow this because the developer using this component is responsible for combining it with an explanatory text next to it.

src/form/NewPasswordInput/PasswordStrength/PasswordStrength.tsx
```diff
@@ -30,10 +30,20 @@ type Props = {
    * @default true
    */
   showMeter?: boolean;
+
+  /**
+   * Override text for the title of the progress bar.
+   */
+  title?: string;
```
gidomanders, line 37: In other components, we combined these properties in a single property called `text` to prevent too many properties when the number of translatable texts increases.

src/table/EpicTable/EpicTable.stories.tsx
```diff
@@ -463,37 +463,37 @@ storiesOf('table/EpicTable', module)
                     {person.firstName}
                   </span>
                 </EpicCell>
-                <EpicCell width={widths.lastName} height={44}>
+                <EpicCell width={widths.lastName} height={44} >
```
gidomanders, line 466: I don't know if these extra spaces are accidental or not, but they don't comply with our code style.

gidomanders (changes requested):

gidomanders (approved):

PR #651, opened 2022-08-04
feature: epic pagination

gidomanders:
Instead of using ellipsis and always display first and last page
numbers, some people prefer the active page to always be in the middle
and display the total number of records.

Added EpicPagination component that works similar to the existing
Pagination component but has a different look and feel.

### DIFF
--- a/src/core/EpicPagination/EpicPagination.scss
+++ b/src/core/EpicPagination/EpicPagination.scss
@@ -1,0 +1,44 @@
+.epic-pagination {
+  user-select: none;
+
+  .page-item {
+    margin: 2px;
+
+    .page-link {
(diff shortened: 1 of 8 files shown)


src/core/EpicPagination/EpicPagination.scss
```diff
@@ -0,0 +1,43 @@
...
+      padding: initial;
+      width: 2.5rem;
+      height: 2.5rem;
+
+      .icon {
+        display: block;
+      }
+    }
+
+    &.disabled .page-link, .page-link.disabled {
```
gidomanders, line 20:
```suggestion
    &.disabled .page-link,
    .page-link.disabled {
```

PR #690, opened 2023-04-19
improvement: form in OpenCloseModal

### LINKED ISSUE #689: OpenCloseModal and forms
We use the OpenCloseModal for forms, as a submit button is displayed in the footer. There is no possibility however to wrap the modal with a form tag, which causes invalid HTML or could cause unexpected behaviour as the submit button could submit another wrapping form as well.

gidomanders:
We use the OpenCloseModal for forms, as a submit button is displayed in the footer. There is no possibility however to wrap the modal with a form tag, which causes invalid HTML or could cause unexpected behaviour as the submit button could submit another wrapping form as well.

Changed OpenCloseModal to use custom modal with form in content wrapping body and footer.

Closes #689


JoelDuurkoop:
Story voor openCloseModal basic redirect als enigste modal voor mij naar [http://localhost:3001/iframe.html?](http://localhost:3001/iframe.html?)
Lijkt niet te gebeuren op https://42bv.github.io/ui/storybook/?path=/docs/core-openclosemodal--basic

